PR #3, opened 2019-07-09
Lighthouse test

styfle:
This improves the lighthouse test

Related to https://github.com/zeit/now-builders/pull/731


test/integration/lighthouse.js
```diff
@@ -1 +1,38 @@
-require('lighthouse');
+const chrome = require('chrome-aws-lambda');
+const { launch } = require('puppeteer-core');
```
guybedford: As described by @lucleray I think because this is a class method, you want to make sure you do `const pupeteer = ` and then call `pupeteer.launch`.

guybedford:
```suggestion
const puppeteer = require('puppeteer-core');
```

test/integration/lighthouse.js
```diff
@@ -1 +1,38 @@
...
+    args: chrome.args,
+    executablePath: await chrome.executablePath,
+    headless: chrome.headless,
+  };
+  return options;
+}
+
+async function getResult(url) {
+  const options = await getOptions();
+  const browser = await launch(options);
```
guybedford:
```suggestion
  const browser = await puppeteer.launch(options);
```

styfle:
I'm not sure if `chrome-aws-lambda` will work in Circle CI since its probably not AWS Lambda.

guybedford:
Did you mean to add `chrome-aws-lambda` as a dependency?

styfle:
It worked!?

styfle:
Oh right, because this doesn't test that the code is executable.

guybedford:
> Oh right, because this doesn't test that the code is executable.

Can you clarify?

guybedford:
(the integration tests are very much executed... try adding a logger into the js file)

lighthouse.js
```diff
@@ -0,0 +1,6 @@
+const nodeFileTrace = require('./src/node-file-trace');
+const entrypoint = '/Users/styfle/Code/zeit/integrations/lighthouse/lighthouse/index.js';
+const workPath = '/Users/styfle/Code/zeit/integrations/lighthouse';
+nodeFileTrace([entrypoint], { base: workPath })
+  .then(({ fileList, warnings }) => console.log(fileList.filter(f => f.includes('lighthouse'))))
+  .catch(console.error);
```
guybedford: I don't think you meant to include this file?

example.js
```diff
@@ -0,0 +1,6 @@
+const nodeFileTrace = require('./src/node-file-trace');
+const entrypoint = '/Users/styfle/Code/foo/now-sharp-example/index.js';
+const workPath = '/Users/styfle/Code/foo/now-sharp-example';
+nodeFileTrace([entrypoint], { base: workPath })
+  .then(({ fileList, warnings }) => console.log(fileList))
+  .catch(console.error);
```
guybedford: Or this one?

styfle:
@guybedford I don't see the `console.log('success')`

lighthouse.js
styfle: Oops, fixed 👍

example.js
styfle: Oops, fixed 👍

test/integration/lighthouse.js
```diff
@@ -1 +1,38 @@
...
+getResult('https://zeit.co/about')
+  .then(result => {
+    if (result && result.lhr && result.lhr.categories) {
+      console.log('success');
+    } else {
+      throw new Error('failed to get result')
+    }
+  })
+  .catch(err => {
+    console.error(err)
```
guybedford: Make this a `throw` to ensure the error exit code propogates to the test. (stderr is permitted due to eg tensorflow warnings)

guybedford:
@styfle yes it's outputting an error - 

> Error: Chromium revision is not downloaded. Run "npm install" or "yarn install".

test/integration/lighthouse.js
styfle: Done

test/integration/lighthouse.js
```diff
@@ -1 +1,38 @@
...
+getResult('https://zeit.co/about')
+  .then(result => {
+    if (result && result.lhr && result.lhr.categories) {
+      console.log('success');
+    } else {
+      throw new Error('failed to get result')
+    }
+  })
+  .catch(err => {
+    throw err;
```
guybedford: I think this actually needs to be `setTimeout(() => throw err)` to avoid the unhandled rejection being swallowed. Or any other approah for the the usual duckery with this.

styfle: Ok good point, fixed

styfle:
I can't think of a way to add proper support here since `chrome-aws-lambda` is the chrome binaries compiled specifically for AWS Lambda.

Should we pursue this or close in favor of https://github.com/zeit/now-builders/pull/731 ??

guybedford:
Agreed if this is lambda specific then let's close this one.

PR #7, opened 2019-07-11
Remove node_modules backtracking filtering

guybedford:
This removes the filtering of asset emission of packages in node_modules of code that lies outside of node_modules.

Having just the `base` filtering is enough I think, although if we relax this emission, I think it is important that we make `base` a mandatory option.

This fixes the prisma issue (unit test derived from that).


guybedford:
Unfortunately, removing this backtracking filter is causing divergent globbing in the azure/cosmos build, blowing up the tests...

guybedford:
The problematic analysis is the following code:

```js
path.resolve(unknown, 'index')
```

which in the analysis will result in globbing `[cwd]/**/index` due to it now knowing anything more about the `unknown` variable.

So the filtering is very much needed to avoid this kind of issue.... although we can see if we can find an alternative way to filter this case while allowing the intentions.

PR #8, opened 2019-07-11
Set default `base` to `process.cwd()`, deprecate `filterBase` option

guybedford:
This makes the `base` option mandatory and deprecates the `filterBase` option as well.

If users really want to have no base, then set `base: '/'` and be aware that you can emit any file on the entire file system. So this doesn't stop that from happening, but just makes it clearer.

By default `base` is set to `process.cwd()`.

Fixes https://github.com/zeit/node-file-trace/issues/6.

### DIFF
--- a/node-file-trace.d.ts
+++ b/node-file-trace.d.ts
@@ -1,6 +1,5 @@
 interface NodeFileTraceOptions {
   base?: string;
-  filterBase?: boolean;
   ignore?: string | string[] | ((path: string) => boolean);
   ts?: boolean;
   log?: boolean;
--- a/src/analyze.js
+++ b/src/analyze.js
@@ -154,7 +154,7 @@ module.exports = async function (id, code, job) {
   const dir = path.dirname(id);
   // if (typeof options.production === 'boolean' && staticProcess.env.NODE_ENV === UNKNOWN)
   //  staticProcess.env.NODE_ENV = options.production ? 'production' : 'dev';
-  cwd = job.base || process.cwd();
+  cwd = job.base;
   const pkgBase = getPackageBase(id);
 
   const emitAssetDirectory = (wildcardPath) => {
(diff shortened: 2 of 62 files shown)


src/analyze.js
```diff
@@ -318,7 +318,7 @@ module.exports = async function (id, code, job) {
     if (!wildcardPattern.endsWith('*'))
       wildcardPattern += '?(' + (job.ts ? '.ts|' : '') + '.js|.json|.node)';
 
-    if (job.ignoreFn(job.base ? path.relative(job.base, wildcardDirPath + wildcardPattern) : wildcardDirPath + wildcardPattern, id))
+    if (job.ignoreFn(path.relative(job.base, wildcardDirPath + wildcardPattern)))
```
styfle:
```suggestion
    if (job.ignoreFn(path.relative(job.base, join(wildcardDirPath, wildcardPattern))))
```

Should this be a path.join() instead of string concat?

node-file-trace.d.ts
```diff
@@ -1,6 +1,5 @@
 interface NodeFileTraceOptions {
   base?: string;
```
styfle:
```suggestion
  base: string;
```

If base is "mandatory" should we remove the optional flag here?

styfle: Or maybe you mean it will have the default if empty so its still optional?

src/analyze.js
guybedford: It doesn't matter since `wildcardDirPath` has already been through `path.resolve`.

node-file-trace.d.ts
guybedford: Exactly, it's mandatory but has a default!

src/analyze.js
styfle: What about Windows support?

guybedford: `path.relative` would effectively normalize that out for us.

PR #15, opened 2019-07-12
Fix require symlinks for yarn workspaces

### LINKED ISSUE #14: Failing to trace node_modules containing symlinks (yarn workspaces)
The issue is that `node-file-trace` is following the symlinks is `node_modules` instead of returning them in the `fileList` output.

## Steps to reproduce

```sh
git clone https://github.com/mattste/now-node-builder-issue
cd now-node-builder-issue
yarn install
cat > build.js <<- "EOF"
const nodeFileTrace = require('@zeit/node-file-trace');

nodeFileTrace(['project-a/index.js'], {
  base: process.cwd(),
  filterBase: true,
  ts: true,
}).then(({fileList}) => {
  console.log(fileList.join('\n'));
});

EOF

node build.js
```

### Actual

```
        project-a/index.js
        project-b/index.js
```

### Expected

```
        project-a/index.js
        project-b/index.js
        node_modules/project-a/index.js
```

cc @guybedford

styfle:
Fixes #14


src/resolve-dependency.js
```diff
@@ -11,7 +11,22 @@ module.exports = function resolveDependency (specifier, parent, job) {
...
   if (resolved.startsWith('node:')) return resolved;
-  return fs.realpathSync(resolved);
+
+  const real = fs.realpathSync(resolved);
+  const original = resolved;
+  let isLink = false;
+  
+  if (resolved !== real) {
+    while (resolved !== '/') {
+      // walk up the tree to confirm that this is a symlink
```
lucleray: I think we can save some `lstat` executions here and just compare `resolve` and `real` to extract their "common part", and look for a symlink there.

styfle: We can probably assume any difference here is a symlink and emit so I pushed a change.

styfle:
@guybedford What am I missing here?

src/resolve-dependency.js
```diff
@@ -11,7 +11,12 @@ module.exports = function resolveDependency (specifier, parent, job) {
   else
     resolved = resolvePackage(specifier, parent, job);
   if (resolved.startsWith('node:')) return resolved;
-  return fs.realpathSync(resolved);
+  const real = fs.realpathSync(resolved);
+  if (resolved !== real) {
+    // this is likely a symlink
+    job.emitFile(resolved, 'symlink', parent);
```
lucleray: If you emit the path including the symlink, you probably don't need to also emit the real path (because the real path is not required).

styfle: I don't understand.

Both the symlink and the real file need to be emitted because the symlink is just a pointer to the real file. Both are necessary.

test/unit/fs-emission-symlink/input.js
```diff
@@ -0,0 +1,3 @@
+const fs = require('fs');
+const { join } = require('path');
+fs.readFile(join(__dirname, 'symlink'));
```
guybedford: I think we're looking to capture the `require` test here rather than an asset emission:

```js
require('./symlink')
```

where `symlink` is a symlink to another CommonJS file

src/resolve-dependency.js
lucleray: For example, you could have this path (not real) `/symlink-folder/file` and the real path is `/folder/file`.

If you add `/symlink-folder/file` to the file list, you're not actually listing anything real (not even a file that contains a pointer). 

That means that:
1. the consumer of the list (eg. `@now/node`) would need to follow symlinks when copying files
2. the consumer of the list doesn't need to copy the real file anymore
3. you avoid having to detect or copy symlinks. You'll just create real files instead of "virtual" ones when you copy.

test/unit/fs-emission-symlink/input.js
styfle: You’re absolutely right!

I will fix the test tomorrow unless you want to do it before then 🙏

styfle: I tried to edit on my phone, we’ll see if that worked 🤞

src/resolve-dependency.js
styfle: I think I see what you mean. My code is immediately emitting the symlink even though there are some cases when the real file will not be emitted (starts with  'node:`, etc).

@guybedford Do you have a solution for this one?

guybedford: It will always be a symlink on this code path because of the early return for the `node:` URL format.

Note that a resolve-dependency is called as part of the emitDependency routine before it itself calls emitFile. Any file emitted by resolve-dependency means that file was necessary for the resolution process, not that that was the resolved file. The emitFile capability was added to resolve-dependency specifically just to emit package.json files (and so this way we don't emit all package.json files, only those that are actually needed by resolution). Emitting package.json + symlinks exactly fits the model.

lucleray: I'm talking about the case where the symlink is directed at a folder and not a file (which is what yarn workspaces is doing). In that case, with the code above, you emit both a "virtual" file and a real file. Is that the expected behaviour? Could you add a test for that?

guybedford: The Node.js resolver as written here will only ever resolve a file, not a folder. The statting specifically checks for file existence.

styfle: I confirmed this fixes yarn workspaces with an integration test here: https://github.com/zeit/now-builders/pull/759

lucleray: Sorry, I explained it wrong. I'm talking about the case where the file itself is not the symlink, but the symlink is located higher in the folder structure. In this case, if we emit the "virtual" file path, we don't need to emit the real file path.

It works but I think we emit some files twice when only one is necessary 🤔

guybedford: Right, that makes complete sense, but this project is just a "tracer" of the file system. It is up to now node to correctly emit the traced filesystem. And yes, it doesn't due to a lack of symlink support both for files and deeper folders as you describe.

PR #18, opened 2019-07-12
node_modules globbing filter test

guybedford:
This includes a test of the node_modules globbing filter, with a fix, while also updating the tests to emit their logging output to keep an eye on the diagnostics here.


src/analyze.js
```diff
@@ -162,7 +162,7 @@ module.exports = async function (id, code, job) {
     const patternPath = wildcardPath.substr(dirIndex);
     const wildcardPattern = patternPath.replace(wildcardRegEx, (_match, index) => {
       return patternPath[index - 1] === path.sep ? '**/*' : '*';
-    }) || '/**/*';
+    }).replace(/([\/\\]\*\*[\/\\]\*)+/g, '/**/*') || '/**/*';
```
styfle: Should this regex be in a variable like you did with `wildcardRegEx`?

guybedford: Sure, I've separated that out.

PR #19, opened 2019-07-13
Add log option to types

guybedford:
This adds the log option to the typing of the library.


styfle:
At some point, I would like to change the source code to typescript so we don't have  mismatches like this.

PR #21, opened 2019-07-16
Emit symlinks directly

guybedford:
This implements an internal realpath routine, which then emits symlinks as direct references.

This way "virtual" files within symlinked folders will not ever be emitted - rather the symlink folder itself is emitted, and then the files are emitted at their realpath.

On the consumer side, a direct lstat applied to the `fileList` can determine if a file is a symlink, and in those cases emit the directory or file symlink as appropriate, but that is the only check needed.

This should allow for a straightforward fix to zeit/now-builders#765.


styfle:
I think this is what @lucleray was talking about: Only emitting the symlink directory.

I am testing with the PR and so far so good https://github.com/zeit/now-builders/pull/765

PR #22, opened 2019-07-16
Use statting instead of read in resolver

guybedford:
This changes the resolver FS to use direct stat calls instead of going straight to a `read` call. Now that statting is part of the base FS API this makes more sense.


styfle:
@guybedford Can you add a test for this?

guybedford:
@styfle I tried but it's actually really hard to test... so considering it as more of a refactoring.

PR #23, opened 2019-07-16
Refactor realpath handling

guybedford:
This refactors the realpath handling to apply to all file emissions (including assets).

This is done by lowering the realpath operation onto `job`, including the way it emits the symlink junctions.


styfle:
@guybedford How about tests?

guybedford:
Sure, I've included a test.

lucleray:
@guybedford Have you considered doing `fs.realpath` on the path, and if realpath <> path, find their common part, which would tell you where the symlink is? Maybe your method is more reliable and faster?

guybedford:
Yes that sounds like it would work. The benefit of this approach is being
able the share the real path cache though.

PR #26, opened 2019-07-22
Add support for mixed ESM and CJS via `mixedModules` flag

styfle:
This adds support for mixing ESM and CJS so that we have feature parity with `ncc`, which is basically `webpack`.

This will fix all the issues that were reported in the latest release:

- https://spectrum.chat/zeit/general/now-deploys-suddenly-not-working~a052cb8f-e95e-4eef-9e81-5e799d1e5ef4
- https://spectrum.chat/zeit/now/did-you-forget-to-add-it-to-dependencies-in-package-json~6b370740-866c-4404-bdcc-cc51a3bc9aec
- https://spectrum.chat/zeit/now/cannot-build-ts-api-with-now-node-anymore~a24f6eb3-201a-4ecb-8688-94b4a56d4abe
- https://github.com/zeit/now-builders/issues/791

## Update

I spoke with Timer and it sounds like Next.js might want to follow the Node.js behavior. If that's the case maybe we don't want this PR to land. We need to discuss with the whole team.

Perhaps we should enable backwards compatibility through a `config` flag like `mixedModules: true`. That way user's who have broken deployments can opt-in to this behavior.

### DIFF
--- a/test/unit/require-esm/output.js
+++ b/test/unit/require-esm/output.js
@@ -1,3 +1,0 @@
-[
-  "test/unit/require-esm/input.js"
-]
--- a/node-file-trace.d.ts
+++ b/node-file-trace.d.ts
@@ -3,6 +3,7 @@ interface NodeFileTraceOptions {
   ignore?: string | string[] | ((path: string) => boolean);
   ts?: boolean;
   log?: boolean;
+  mixedModules?: boolean;
   readFile?: (path: string) => Buffer | string | null;
   stat?: (path: string) => Object | null;
   readlink?: (path: string) => string | null;
--- a/src/analyze.js
+++ b/src/analyze.js
@@ -222,7 +222,7 @@ module.exports = async function (id, code, job) {
     }
   });
 
-  if (!isESM) {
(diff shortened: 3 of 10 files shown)


test/unit/require-esm/output.js
```diff
@@ -1,3 +0,0 @@
-[
-  "test/unit/require-esm/input.js"
-]
```
styfle: I deleted this test because it was contrary to the new behavior

styfle:
@Timer I just confirmed that `ncc` works with the unit test I added.

```
cd test/unit/mixed-esm-cjs
ncc build input.js
node dist
```

This change is necessary to maintain backwards compatibility. I would be open to using a flag to opt-in to this non-standard behavior if you prefer.

Timer:
Ah, this is because `ncc` uses webpack's legacy `javascript/auto` source handling method instead of `javascript/esm` or `javascript/dynamic`.

styfle:
@Timer I added a flag `mixedModules` so that we can opt-in to this legacy behavior.

PR #28, opened 2019-07-29
Fix photon binary resolution

timsuchanek:
We renamed the binary of photon from `prisma` to `query-engine-${PLATFORM}`.
In order to support that, I changed the special case to copy the whole runtime directory.


test/unit/prisma-photon/actual.js
```diff
@@ -0,0 +1,4 @@
+[
+  "test/unit/prisma-photon/input.js",
+  "test/unit/prisma-photon/node_modules/@generated/photon/index.js"
+]
```
styfle: This file shouldn't be committed

test/unit/prisma-photon/node_modules/@generated/photon/runtime/prisma
```diff
@@ -1 +0,0 @@
-asdf
```
styfle: Perhaps you should rename this file from `prisma` to `query-engine-darwin`

test/unit/prisma-photon/actual.js
timsuchanek: Thanks! `actual.js` should then probably be added to `.gitignore`.

test/unit/prisma-photon/node_modules/@generated/photon/runtime/prisma
timsuchanek: Yep, as node_modules are gitignored, I had to force add it.

PR #33, opened 2019-08-02
Update azure-pipelines.yml to fix Windows CI

### LINKED ISSUE #31: Needs CI for Windows
To prevent regressions in `now dev` on Windows (i.e. #30).

Recommend [Azure Pipelines](https://azure.microsoft.com/en-us/services/devops/pipelines/).

styfle:
Previously, Azure Pipline was not running tests. It would only run `yarn install`. 

This PR fixes that so the tests are also run.

Fixes #31

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 dist
 coverage
 test/**/dist
+test/**/actual.js
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -16,8 +16,5 @@ steps:
     versionSpec: '10.x'
   displayName: 'Install Node.js'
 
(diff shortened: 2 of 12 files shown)


styfle:
@TooTallNate I decided to remove the `loopback.js` integration test for Windows just to get these passing. It was failing with `Error: Cannot find module 'serve-favicon'`.

I think it would be better to add an integration test with the following to narrow it down:

```js
var favicon = require('serve-favicon');
var path = require('path');

module.exports = function(icon, options) {
  icon = icon || path.join(__dirname, '../../favicon.ico');
  return favicon(icon, options);
};
```

But I'll do that in a new PR.

PR #40, opened 2019-08-09
Trace `.tsx` input files

### LINKED ISSUE #39: .tsx files used as an entrypoint for an API fail to get compiled
I thought I was out of the woods with https://github.com/zeit/now-builders/issues/869 getting fixed, but I hit a new problem.

I've got some routes handled by `.tsx` files:

```
{
	"builds": [
		{
			"src": "api-handler.tsx",
			"use": "@now/node"
		}
	],
	"routes": [
		{ "src": "/api-typescript", "dest": "api-handler.tsx" },
	]
}
```

And on the latest node builder they fail to get compiled, resulting in runtime errors due to TypeScript syntax remaining intact.

Repo that reproduces the issue: https://github.com/tylerc/zeit-now-ts-debugging

This is an issue everywhere. On my Windows machine, my Mac, and when deployed via `now` (so must be on Linux as well).

I traced back the cause to `node-file-trace`. I'll be filing a PR shortly.

Pinging @styfle since you were the one who handled my last issue.

tylerc:
This fixes #39 the issue I just filed around `.tsx` files being unable to be used as the entrypoint for a route.

I noticed in `@now/node` this callback was responsible for compiling my `.ts` files, but it was never getting called for my `.tsx` files: https://github.com/zeit/now-builders/blob/canary/packages/now-node/src/index.ts#L177

This pull request makes `node-file-trace` aware that `.tsx` is a valid file extension for TypeScript files.


tylerc:
Pinging @styfle here as well.

src/node-file-trace.js
```diff
@@ -27,7 +27,7 @@ module.exports = async function (files, opts = {}) {
   await Promise.all(files.map(file => {
     const path = resolve(file);
     job.emitFile(job.realpath(path), 'initial');
-    if (path.endsWith('.js') || path.endsWith('.node') || job.ts && path.endsWith('.ts'))
+    if (path.endsWith('.js') || path.endsWith('.node') || job.ts && (path.endsWith('.ts') || path.endsWith('.tsx')))
```
styfle: Thanks for finding and fixing this issue! Can you add a unit test for this?

tylerc: Sure! Give me a bit to figure out how your tests are organized and I'll push something up.

tylerc:
@styfle unit tests have been pushed. Since it didn't quite fit the mold of the other tests, I had to add another call to `it()` in `unit.test.js`, hopefully that's okay.

I also found a couple other places that referenced `.ts` but not `.tsx`, so I fixed those up as well.

I couldn't figure out how to test the one in `analyze.js` (nothing ever seemed to get to that code) but both changes in `node-file-trace.js` are covered inthe unit test I added.

test/unit/tsx2/output.js
```diff
@@ -0,0 +1,4 @@
+[
+    "test\\unit\\tsx2\\dep.tsx",
```
styfle:
```suggestion
    "test/unit/tsx2/dep.tsx",
```

test/unit/tsx2/output.js
```diff
@@ -0,0 +1,4 @@
+[
+    "test\\unit\\tsx2\\dep.tsx",
+    "test\\unit\\tsx2\\input.tsx"
```
styfle:
```suggestion
    "test/unit/tsx2/input.tsx"
```

test/unit/tsx2/input.tsx
```diff
@@ -0,0 +1,3 @@
+import * as fs from "fs";
```
styfle: Let's rename this directory from `tsx2` to `tsx-input`.

test/unit.test.js
```diff
@@ -32,6 +32,48 @@ for (const unitTest of fs.readdirSync(join(__dirname, 'unit'))) {
     console.log('skipping symlink test on Windows: ' + unitTest);
     continue;
   }
+  if (unitTest === 'tsx2') {
```
styfle: As far as I can tell, the only difference between this unit test and the others is that there is a `readFileMock`.

So instead of copy/paste, let's keep it [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and instead add a `expect(readFileMock.mock.calls.length).toBe(2);` when the unitTest is the tsx one.

Does that make sense?

test/unit/tsx2/input.tsx
tylerc: Will do!

test/unit.test.js
tylerc: Yes, it does. I'll make that change and push it up.

tylerc:
@styfle All your suggestions should now be integrated.

tylerc:
Thank you!

PR #51, opened 2019-08-18
Support webpack-style Node requires

Timer:
This patches multiple bugs with our static evaluator:
- `MemberExpression` did not handle `computed: true` properties when encountered
- `BinaryExpression` returned wrong value for `A || UNKNOWN -> A if A is truthy` case
- `ObjectExpression` did not wrap computed value in `{ value: ... }` for `walk` compatibility

The sum of these changes allow us to support webpack-style `target: 'node'` emitted `require`s!

### DIFF
--- a/src/utils/static-eval.js
+++ b/src/utils/static-eval.js
@@ -260,7 +260,17 @@ const visitors = {
     if (node.property.type === 'Identifier') {
       if (typeof obj.value === 'object' && obj.value !== null) {
         if (node.computed) {
-          return { value: WILDCARD, wildcards: [node] };
+          // See if we can compute the computed property
+          const computedProp = walk(node.property);
+          if (computedProp && computedProp.value) {
+            const val = obj.value[computedProp.value];
+            if (val === UNKNOWN) return;
+            return { value: v };
+          }
+          // Special case for empty object
+          if (!obj.value[UNKNOWN] && Object.keys(obj).length === 0) {
+            return { value: undefined };
+          }
         }
         else if (node.property.name in obj.value) {
           const val = obj.value[node.property.name];
--- a/src/utils/static-eval.js
+++ b/src/utils/static-eval.js
@@ -69,7 +69,7 @@ const visitors = {
       }
       // A || UNKNOWN -> A if A is truthy
       if (!('test' in l) && op === '||' && l.value)
-        return node.right;
+        return l;
       return;
     }
 
@@ -142,7 +142,7 @@ const visitors = {
       if (op === '^') return { value: l.value ^ r.value };
       if (op === '&&') return { value: l.value && r.value };
       if (op === '||') return { value: l.value || r.value };
-    }      
+    }
     return;
   },
   CallExpression (node, walk) {
@@ -151,7 +151,7 @@ const visitors = {
     let fn = callee.value;
     if (typeof fn === 'object' && fn !== null) fn = fn[FUNCTION];
     if (typeof fn !== 'function') return;
-    
+
     const ctx = node.callee.object && walk(node.callee.object).value || null;
 
     // we allow one conditional argument to create a conditional expression
@@ -250,12 +250,15 @@ const visitors = {
   },
   MemberExpression (node, walk) {
     const obj = walk(node.object);
-    // do not allow access to methods on Function 
+    // do not allow access to methods on Function
     if (!obj || 'test' in obj || typeof obj.value === 'function')
       return;
     if (node.property.type === 'Identifier') {
       if (typeof obj.value === 'object' && obj.value !== null) {
-        if (node.property.name in obj.value) {
+        if (node.computed) {
+          return { value: WILDCARD, wildcards: [node] };
+        }
+        else if (node.property.name in obj.value) {
           const val = obj.value[node.property.name];
           if (val === UNKNOWN)
             return;
@@ -297,7 +300,7 @@ const visitors = {
       if (value.value === UNKNOWN) return;
       obj[keyValue.value] = value.value;
     }
-    return obj;
+    return { value: obj };
   },
   TemplateLiteral (node, walk) {
     let val = { value: '' };
--- a/src/utils/static-eval.js
+++ b/src/utils/static-eval.js
@@ -265,7 +265,7 @@ const visitors = {
           if (computedProp && computedProp.value) {
             const val = obj.value[computedProp.value];
             if (val === UNKNOWN) return;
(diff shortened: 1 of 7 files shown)


src/utils/static-eval.js
```diff
@@ -297,7 +300,7 @@ const visitors = {
       if (value.value === UNKNOWN) return;
       obj[keyValue.value] = value.value;
     }
-    return obj;
+    return { value: obj };
```
guybedford: Great catch.

src/utils/static-eval.js
```diff
@@ -69,7 +69,7 @@ const visitors = {
       }
       // A || UNKNOWN -> A if A is truthy
       if (!('test' in l) && op === '||' && l.value)
-        return node.right;
+        return l;
```
guybedford: This fix is correct, but I'm not sure I follow how it makes the test case work since `({}[chunkId] || chunkId)` should effectively be a case of `undefined || UNKNOWN` (or possibly `UNKNOWN || UNKNOWN` if we want to support prototype properties).

The `undefined || UNKNOWN` case we can handle, the `UNKNOWN || UNKNOWN` case would likely require us to distinguish between a `UNKNOWN_WITH_SIDE_EFFECT` and `UNKNOWN_PURE`. That seems the ideal analysis fix, but I would suggest perhaps just going with `undefined || UNKNOWN` for now. This case should fall out of the standard logical analysis fine though I believe (as opposed to this path which is somewhat of a shortcut inlining the fact that the `UNKNOWN` statement in `A || UNKNOWN` is side-effect free by definition).

So that means that we need to ensure the analysis is returning `undefined` for `{}[chunkId]` (and not `UNKNOWN` or `WILDCARD` as it currently does in this PR).

src/utils/static-eval.js
```diff
@@ -250,12 +250,15 @@ const visitors = {
...
-    // do not allow access to methods on Function 
+    // do not allow access to methods on Function
     if (!obj || 'test' in obj || typeof obj.value === 'function')
       return;
     if (node.property.type === 'Identifier') {
       if (typeof obj.value === 'object' && obj.value !== null) {
-        if (node.property.name in obj.value) {
+        if (node.computed) {
+          return { value: WILDCARD, wildcards: [node] };
+        }
```
guybedford: As mentioned above, our goal here should be to get `{}[chunkId]` to evaluate to `undefined`.

Note also that the `WILDCARD` assumption assumes the result to be a string (`WILDCARD` is a string value, used in string concatentation to expand into the globs).

If we were analyzing a member expression that didn't result in a string value, then this would be wrong.

Ideally as mentioned above we could have some `UNKNOWN_WITH_SIDE_EFFECT` and `UNKNOWN_PURE` analysis going on here, but since we don't it is probably best to explicitly handle the `obj[unknownIdentifier]` case directly here for an empty object, along with proper support for computed properties while we are about it.

Something like the following should work - 

```js
if (node.computed) {
  const computedProp = walk(node.property);
  // proper computed property support, because we can and we're here anyway
  if (computedProp && computedProp.value) {
    const val = obj.value[computedProp.value];
    if (val === UNKNOWN) return;
    return { value: v };
  }
  // empty object case for this test
  if (!obj.value[UNKNOWN] && Object.keys(obj).length === 0) {
    return { value: undefined };
  }
}
```

src/utils/static-eval.js
```diff
@@ -151,7 +151,7 @@ const visitors = {
     let fn = callee.value;
     if (typeof fn === 'object' && fn !== null) fn = fn[FUNCTION];
     if (typeof fn !== 'function') return;
-    
+
     const ctx = node.callee.object && walk(node.callee.object).value || null;
```
guybedford: I just noticed that this `walk(node.callee.object).value` case would throw if there was an undefined return from walk. That would throw the entire build, so probably worth fixing this.

src/utils/static-eval.js
guybedford: I realise that the ideas of pure and not pure are actually ncc-specific, since ncc used to substitute into the expressions (requiring that this analysis is pure and substitutable). So lifting that constraint may mean we can make more analysis optimizations here that ncc can't. Given that you can also ignore my `UNKNOWN_PURE` and `UNKNOWN_WITH_SIDE_EFFECT` comments :)

Timer: Thanks, I wasn't aware that `{ value: undefined }` had similar semantics to wildcard! Maybe you can explain the 4 to me on Slack one day (wildcard, unknown, { value: undefined }, undefined).

src/utils/static-eval.js
Timer: In this case it was making it work by returning the `WILDCARD` for the visited member expression on the left hand side (`{}[unknownComputedProp]`), but as you said, we don't want to use `WILDCARD`!

Either way, this fix is still correct! 😄

src/utils/static-eval.js
Timer: Fixed this, I think I encountered this bug at some point while I was working with the `WILDCARD` implementations but chalked it off as something I did wrong.

src/utils/static-eval.js
guybedford: `{ value: undefined }` is exactly an undefined value in the analysis, because `{}.x === undefined` in JS is what we want. `UNKNOWN` is mainly used for input values into the analysis boundary through the variables and functions. Otherwise all analysis values are wrapped up in `{ value: ... }` the `undefined` means no analysis available, but note that many cases that return `undefined` can probably be upgraded to return a new `UNKNOWN` analysis value since we are no longer tracking side effects. That is it may make sense to extend `UNKNOWN` from beyond just the analysis inputs. But keeping an eye on performance would be important too as these are super hot paths.

src/utils/static-eval.js
guybedford: Great.

src/utils/static-eval.js
```diff
@@ -250,12 +254,25 @@ const visitors = {
...
+          // See if we can compute the computed property
+          const computedProp = walk(node.property);
+          if (computedProp && computedProp.value) {
+            const val = obj.value[computedProp.value];
+            if (val === UNKNOWN) return;
+            return { value: val };
+          }
+          // Special case for empty object
+          if (!obj.value[UNKNOWN] && Object.keys(obj).length === 0) {
+            return { value: undefined };
```
guybedford: Very glad to hear it's still passing under this comprehensive approach.

PR #57, opened 2019-08-30
Update typescript types

AndyBitz:
Updates the types for typescript.

Is required for https://github.com/zeit/now/pull/2926

### DIFF
--- a/node-file-trace.d.ts
+++ b/node-file-trace.d.ts
@@ -51,7 +51,7 @@ export interface NodeFileTraceResult {
   reasons: NodeFileTraceReasons;
 }
 
-export = function NodeFileTrace(
+export default function NodeFileTrace(
   files: string[],
   opts: NodeFileTraceOptions
 ): Promise<NodeFileTraceResult>;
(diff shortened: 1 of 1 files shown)


node-file-trace.d.ts
```diff
@@ -1,14 +1,57 @@
...
+    type: string;
+    ignored: boolean;
+    parents: string[];
+  };
+}
+
+export interface NodeFileTraceResult {
+  fileList: string[];
+  esmFileList: string[];
+  reasons: NodeFileTraceReasons;
```
styfle: You're missing `warnings: Error[];` here

node-file-trace.d.ts
```diff
@@ -1,14 +1,57 @@
...
+}
 
-export = NodeFileTrace;
```
styfle: Why was this changed to `default`?

AndyBitz: I've used `export = ` before, then linked it and typescript complained when I've tried to build a package

styfle: I believe `esModuleInterop` should take care of it (I think it supports both syntax options to export). But if it was failing in `@now/next`, then we can change it.

styfle:
@AndyBitz Looks good! Feel free to release (probably semver patch?) after CI passes.

node-file-trace.d.ts
AndyBitz: yup, it failed in `@now/next`

PR #68, opened 2019-09-11
Add the kangax ecmascript feature test suite

styfle:
This adds a massive [test suite](https://github.com/kangax/compat-table) that powers [https://kangax.github.io](https://kangax.github.io/compat-table/es6/) to make sure we can upgrade `acorn` without regressing in the future.

It also confirms that `try/catch` is not parsed properly as @Timer found last week.

I also added a README so the next person who needs to update the test suite can do so easily.

### DIFF
--- a/test/ecmascript.test.js
+++ b/test/ecmascript.test.js
@@ -7,18 +7,19 @@ const rimraf = require('rimraf');
 const readFile = promisify(fs.readFile);
 const writeFile = promisify(fs.writeFile);
 
+const randomTmpId = Math.random().toString().slice(2);
+const tmpdir = path.resolve(os.tmpdir(), `node-file-trace-ecmascript${randomTmpId}`);
+rimraf.sync(tmpdir);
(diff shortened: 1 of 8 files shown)


test/ecmascript.test.js
```diff
@@ -0,0 +1,64 @@
...
+
+const randomTmpId = Math.random().toString().slice(2);
+const tmpdir = path.resolve(os.tmpdir(), `node-file-trace-ecmascript${randomTmpId}`);
+rimraf.sync(tmpdir);
+fs.mkdirSync(tmpdir);
+console.log('created directory ' + tmpdir);
+
+ // These are tests known to fail so we skip them
+const ignoreCategories = new Set([
+  'optional catch binding',
```
styfle: In a future PR, I will update `acorn` to see if it fixes "optional catch binding".

```
try {
  foo();
} catch {
  // nothing
}
```

PR #79, opened 2019-11-20
Handle babel/ts compiled imports

jkrems:
Fixes https://github.com/zeit/node-file-trace/issues/76


test/unit/asset-fs-inline-path-ts/input.js
```diff
@@ -0,0 +1,22 @@
...
+    if (mod != null) for (var k in mod) if (Object.hasOwnProperty.call(mod, k)) result[k] = mod[k];
+    result["default"] = mod;
+    return result;
+};
+Object.defineProperty(exports, "__esModule", { value: true });
+const fs_1 = __importDefault(require("fs"));
+const path = __importStar(require("path"));
+console.log(fs_1.default.readFileSync(path.join(__dirname, 'asset.txt'), 'utf8'));
+
+/* Input source (with modules: 'commonjs'):
```
styfle: To be clear, these helpers are only generated when `esModuleInterop: true`.

Perhaps there should be one more test for when `esModuleInterop: false` ?

Or, I suppose that TS won't compile if `esModuleInterop: false` in this case?

jkrems: I think the false case already is tested, or at least it should be handled by the `default` key on the fs/path stubs:

```js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
const path = require("path");
const fs_1 = require("fs");
fs_1.default.readFileSync(path.resolve('foo'));
```

jkrems: To immediately correct myself: Doesn't look like there's a test (according to a quick grep) but the code *looks* like it should work.

jkrems: Added that test as well to make sure regressions are caught.

PR #84, opened 2019-11-20
Add node 12 to CI

styfle:
Node 12 is LTS and we are going to begin supporting it in `@now/node` very soon.


.github/main.workflow
```diff
@@ -1 +0,0 @@
-
```
styfle: This was some file left over from legacy GitHub Actions so I deleted it

styfle:
Something looks wrong....the print version step shows 10.17 for both 10 and 12 🤔

Issue #85, opened 2019-11-29
Absolute path mapping


elderapo:
It would be cool to add absolute path mapping support to this module. It is necessary to make typescript paths (`compilerOptions.paths` in `tsconfig.json`) work in `@now/node`.

styfle:
This was implemented in https://github.com/vercel/nft/pull/90

PR #90, opened 2020-01-03
Paths implementation

guybedford:
This implements a simple paths implementation for easier integration into other workflows.

We could also implement a full resolve hook option here as well, but this seemed somewhat simpler for now.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -42,6 +42,20 @@ const { fileList } = await nodeFileTrace(files, {
 
 Any files/folders above the `base` are ignored in the listing and analysis.
 
+#### Paths
+
+Custom resolution path definitions to use.
+
+```js
+const { fileList } = await nodeFileTrace(files, {
+  paths: {
+    'utils/': '/path/to/utils/'
+  }
+});
+```
+
+Trailing slashes map directories, exact paths map exact only.
+
 #### TypeScript
 
 Both JavaScript and TypeScript source files can be traced.
--- a/src/node-file-trace.js
+++ b/src/node-file-trace.js
@@ -42,6 +42,7 @@ module.exports = async function (files, opts = {}) {
 class Job {
   constructor ({
     base = process.cwd(),
+    paths = {},
     ignore,
     log = false,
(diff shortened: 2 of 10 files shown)


guybedford:
Just working on adjustment to this to ensure compatibility with node_modules resolution.

guybedford:
Directly supporting tsconfig.json `baseUrl` configuration could be an alternative but I must admit I'm not a fan of this resolution mechanism as no runtimes really support this sort of resolution apart from just the old AMD loaders.

Paths as written here more closely follows the import maps style of declaration, as opposed to the older AMD approach. I've also moved the paths application to happen after node_modules resolution so it behaves as a fallback for compatibility.

An alternative could be to make this configuration exactly like the ts baseUrl configuration, and then later as a follow up that could enable automatically inferring this.

I'm open to either approach though.

styfle:
There is a longstanding feature request (even back from ncc) to support TS `paths` (#85).

Even if it's non-standard, we should provide a way for `@now/node` to convert TS paths to NFT paths.

The alternative is waiting another year or more to TS to support import maps (I'm not even sure if Node.js supports import maps yet).

guybedford:
@styfle it's no problem if we want to change it to align with ts paths. The main priority was to get the simplest approach working for the API usage (and avoid forking there), but I'm certainly not tied to this API.

styfle:
@guybedford I like the idea of aligning NFT with Node, but I'm not sure how to convert TS [paths](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping) to this format considering TS supports an array on the right hand side.

```js
{
  "compilerOptions": {
     // This must be specified if "paths" is
    "baseUrl": ".",
    "paths": {
       // This mapping is relative to "baseUrl" and can have fallbacks
      "jquery": ["node_modules/jquery/slim", "node_modules/jquery/all"]
    }
  }
}
```

How would you write that with NFT?

guybedford:
@styfle supporting array paths here sounds like a worthy feature to me.

Issue #25, opened 2019-07-22
Cache trace results for paths which start with node_modules


ywg-jean:
With the recent update, the @now/node builder switched from ncc to node-file-trace. This has already improved the build performance for my endpoints. 
zeit/now-builders#786 illustrated that of the remaining time, most was spent in node-file-trace either the trace or the typescript compilation. A manual check running only node-file-trace demonstrated that the trace was the greatest contributor. 

While the time is perfectly acceptable for a first time compilation, repeated compilation of the same endpoint in `now dev` with trival changes (add a comment to the file, add a console log etc) takes the same time. 
This leads me to believe that node-file-trace doesn't cache trace results from one run to another. 

The attached project  [node-20.zip](https://github.com/zeit/node-file-trace/files/3418943/node-20.zip)(initially created by @styfle updated to the last versions of the stack tools) exhibits this behaviour. 

I suggest updating node-file-trace to keep a cache of the trace result for each path in the result (this could take the form of a dictionnary with a path as key and an object with the hash of the content at that path associated with the resolved list of dependencies as value), retracing a codebase which was previously traced would be able to avoid retracing all the files which have already been traced when an entry is found in the map and the content hash matches. If the content hash doesn't match, the entry must be discarded and retracing must proceed on all the file's dependencies.

styfle:
Hi @ywg-jean 

There actually is a caching mechanism in the `@now/node` builder.

https://github.com/zeit/now-builders/blob/canary/packages/now-node/src/index.ts#L87-L89

They just don't persist between each build because the build() function gets invoked when one of the traced/watched files changes.

I'm not entirely sure that what you suggest could even be implemented without introducing bugs (when does the cache get invalidated, for example?)

wtgtybhertgeghgtwtg:
It'd be more than a little work, but maybe something that does a check depending on the version, and never automatically invalidates?
```js
// If this resolves to a version of `some-module` that's been seen before, the list of files in `some-module` that'd be positive is cached, as well as a list of transitive dependencies that `some-module` uses and from where they'd be resolved.
// This'd mean only the `package.json` of `some-module` would have to be touched in cached cases, and the cache would be effective regardless of what the transitive dependencies would resolve to.
import {something} from 'some-module'; 
```

styfle:
Since `vc dev` no longer uses node-file-trace, I don't think this feature is necessary.

[...]

PR #94, opened 2020-01-03
Implement simple cache persistence between traces

guybedford:
This provides a `cache` option which stores the stat, symlink and file cache between runs for reuse when the file system is known to not have any changes between runs.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -68,6 +68,21 @@ Ignore will also accept a function or globs.
 
 Note that the path provided to ignore is relative to `base`.
 
+#### Cache
+
+To persist the file cache between builds, pass an empty `cache` object:
+
+```js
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -68,6 +68,21 @@ Ignore will also accept a function or globs.
...
+```js
+const cache = Object.create(null);
+const { fileList } = await nodeFileTrace(['index.ts'], { cache });
+// later:
+{
+  const { fileList } = await nodeFileTrace(['index.ts'], { cache });
+}
+```
+
+Note that cache invalidations are not supported so the assumption is that the file system is not changed between runs.
```
Timer: We should bold this.

styfle:
Does this fix #25?

PR #98, opened 2020-01-13
Fix `passport-trakt` recursive resolve

guybedford:
Fixes https://github.com/zeit/node-file-trace/issues/95 which was caused by an error in the `pkginfo` package.json lookup inlining logic.

### DIFF
--- a/src/analyze.js
+++ b/src/analyze.js
@@ -566,7 +566,7 @@ module.exports = async function (id, code, job) {
               let pjsonPath = path.resolve(id, '../package.json');
               const rootPjson = path.resolve('/package.json');
               while (pjsonPath !== rootPjson && !existsSync(pjsonPath))
-                pjsonPath = path.resolve(pjsonPath, '../package.json');
+                pjsonPath = path.resolve(pjsonPath, '../../package.json');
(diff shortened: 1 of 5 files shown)


src/analyze.js
```diff
@@ -566,7 +566,7 @@ module.exports = async function (id, code, job) {
               let pjsonPath = path.resolve(id, '../package.json');
               const rootPjson = path.resolve('/package.json');
               while (pjsonPath !== rootPjson && !existsSync(pjsonPath))
-                pjsonPath = path.resolve(pjsonPath, '../package.json');
+                pjsonPath = path.resolve(pjsonPath, '../../package.json');
```
styfle: I'm surprised this works. Why would you jump up two directories each loop instead of one at a time?

guybedford: It's only going up one directory, the first `../` gets to the current directory, and the second one goes up to the next directory.

Timer: This bug, illustrated:
```
> path.resolve('/foo/bla.txt', '..')
'/foo'
> path.resolve('/foo/bla.txt', '../bla.txt')
'/foo/bla.txt'
```

PR #99, opened 2020-01-15
Fixup static test case

guybedford:
This adds a test case for the previous error on acorn-stage3, while removing the dependence on acorn-stage3 and instead inlining the syntax plugins.

Note that this plugin list / exact Acorn config does need to be carefully maintained over time.

Fixes https://github.com/zeit/node-file-trace/issues/55.


src/analyze.js
```diff
@@ -12,8 +12,17 @@ const { pregyp, nbind } = require('./utils/binary-locators');
 const interopRequire = require('./utils/interop-require');
 const handleSpecialCases = require('./utils/special-cases');
 const resolve = require('./resolve-dependency.js');
-const stage3 = require('acorn-stage3');
-acorn = acorn.Parser.extend(stage3);
+
+// Note: these should be deprecated over time as they ship in Acorn core
```
styfle: Nice!

src/utils/acorn-static-class-features.js
```diff
@@ -0,0 +1,81 @@
...
+    this._inStaticFieldValue = true
+    field.value = this.parseExpression()
+    this._inStaticFieldValue = oldInFieldValue
+  } else field.value = null
+}
+
+const privateClassElements = require("acorn-private-class-elements")
+
+module.exports = function(Parser) {
+  const ExtendedParser = privateClassElements(Parser)
```
styfle: I'm confused why this plugin was added?

Is this just a patched version of an existing plugin?
Or did it never exist before?

guybedford: Originally I must admit I had no idea why this PR was working. It seems like this fix is actually due to the latest commit on master at https://github.com/acornjs/acorn-static-class-features/commit/e7295fc9ea64d46f166f53c757884dd9b0d1e177 which isn't currently released.

guybedford:
@styfle sure I've added the fast-glob integration test here and it's definitely working now.

leerob:
Any ETA when this will be released to address https://github.com/zeit/node-file-trace/issues/55? 🙏

styfle:
There's a PR to bump nft in `@now/node` here https://github.com/zeit/now/pull/3613

PR #106, opened 2020-03-27
Fix Node 12 CI (from ncc)

guybedford:
This ports the 12 CI fixes from https://github.com/zeit/ncc/pull/525 and bumps dependencies

### DIFF
--- a/prepare-install.js
+++ b/prepare-install.js
@@ -1,16 +1,13 @@
 const { unlinkSync, readFileSync, writeFileSync } = require('fs');
 const { join } = require('path');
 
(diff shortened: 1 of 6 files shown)


prepare-install.js
```diff
@@ -8,9 +8,6 @@ const isNode12 = process.version.startsWith('v12.');
 
 if (isWin || isNode12) {
```
styfle: I think you can remove the isNode12 check here and only delete yarn.lock for windows

guybedford: Added.

PR #114, opened 2020-05-19
Test `process.cwd()`

styfle:


guybedford:
@styfle `base` could be a sensible option, although note that it restricts files below it. A custom option for `processCwd` could be useful too.

styfle:
I think the root cause was mistaken. We are actually hitting an issue with webpack outputs not including assets similar to #111.

PR #120, opened 2020-06-09
Support a processCwd option

guybedford:
This adds support for a `processCwd` configuration option, including updating the current Webpack test case to use a `path.join(process.cwd()` expression to ensure this is correctly locating assets.

### DIFF
--- a/src/node-file-trace.js
+++ b/src/node-file-trace.js
@@ -42,6 +42,7 @@ module.exports = async function (files, opts = {}) {
 class Job {
   constructor ({
     base = process.cwd(),
+    processCwd,
     paths = {},
     ignore,
     log = false,
@@ -71,6 +72,7 @@ class Job {
       }
     }
     this.base = base;
+    this.cwd = resolve(processCwd || base);
     const resolvedPaths = {};
     for (const path of Object.keys(paths)) {
       const trailer = paths[path].endsWith('/');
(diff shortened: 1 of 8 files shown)


src/node-file-trace.js
```diff
@@ -71,6 +72,7 @@ class Job {
       }
     }
     this.base = base;
+    this.cwd = resolve(processCwd || base);
```
styfle, line 75: Where is `this.cwd` used? I don't see it 🤔

styfle: I only see `job.cwd` above but that cant be the only change right?

test/unit.test.js
```diff
@@ -51,6 +51,7 @@ for (const unitTest of fs.readdirSync(join(__dirname, 'unit'))) {
 
     const { fileList, reasons } = await nodeFileTrace([join(unitPath, inputFileName)], {
       base: `${__dirname}/../`,
+      processCwd: unitPath,
```
styfle, line 54: I was expecting the base and processCwd to be the same here. Why is it different?

styfle: I suppose base is the entire `test` directory where the process cwd is the `test/unit/<fixture>` so this seems correct 👍

styfle: Let's also add `processCwd` to the `integration.test.js` too

src/node-file-trace.js
guybedford: `this.cwd` is `job.cwd`

styfle: Indeed. I was expecting more changes than this one line 😅 

Something is still wrong though as I mentioned with the Next.js example in slack.

test/unit.test.js
guybedford: Ok, done.

PR #121, opened 2020-06-11
Remove invalid scope modification

guybedford:
This fixes the bug introduced with this scope modification. I mixed up `arguments` and `params`, thinking this was a function declaration / expression case where the replacement modifies the scope but obviously `params` is a call expression where the param replacement does not affect the scope declarations at all.

A test for this would check that `f(__webpack_require(4))` does not attempt to alter the scope. But that this was even attempted in the first place was completely invalid behaviour so I don't see any reason to have an explicit test case for what is effectively a completely arbitrary bug case. But happy to put one together if you would still prefer @styfle.


src/utils/wrappers.js
```diff
@@ -390,7 +390,6 @@ function handleWrappers (ast) {
                     parent.callee = replacement;
                   }
                   else if (parent.arguments && parent.arguments.some(arg => arg === node)) {
-                    delete parent.scope.declarations[node.id];
```
styfle, line 393: We need a test for this code path because it must not have been tested or else we would have caught the bug

guybedford: Ok, added.

guybedford:
@styfle the invariant is that any FunctionExpression or FunctionDeclaration will have a `scope` property attached to it, and that the `scope` property will always have a `declarations` object attached to it. If we had a suitable typing system for ASTs it would show this. In both of those scenarios one can see the `FunctionExpression` check that confirms the invariant holds directly.

styfle:
> suitable typing system for ASTs

Maybe we can look into using typescript for this?

PR #128, opened 2020-06-26
Fix trailing slash handling

guybedford:
This fixes up the resolution of specifiers with a trailing slash to ensure they skip file resolution when there exists a file by the same name.

Fixes https://github.com/vercel/node-file-trace/issues/127

### DIFF
--- a/src/resolve-dependency.js
+++ b/src/resolve-dependency.js
@@ -21,7 +21,7 @@ function resolvePath (path, parent, job) {
 }
 
 function resolveFile (path, parent, job) {
-  if (path[path.length - 1] === '/') return;
+  if (path.endsWith('/')) return;
   path = job.realpath(path, parent);
   if (job.isFile(path)) return path;
   if (job.ts && path.startsWith(job.base) && path.substr(job.base.length).indexOf(sep + 'node_modules' + sep) === -1 && job.isFile(path + '.ts')) return path + '.ts';
--- a/src/resolve-dependency.js
+++ b/src/resolve-dependency.js
@@ -5,10 +5,13 @@ const { isAbsolute, resolve, sep } = require('path');
 // (package.json files are emitted as they are hit)
 module.exports = function resolveDependency (specifier, parent, job) {
   let resolved;
-  if (isAbsolute(specifier) || specifier === '.' || specifier === '..' || specifier.startsWith('./') || specifier.startsWith('../'))
-    resolved = resolvePath(resolve(parent, '..', specifier), parent, job);
-  else
+  if (isAbsolute(specifier) || specifier === '.' || specifier === '..' || specifier.startsWith('./') || specifier.startsWith('../')) {
+    const trailingSlash = specifier[specifier.length - 1] === '/' || specifier[specifier.lenggth - 1] === '\\';
+    resolved = resolvePath(resolve(parent, '..', specifier) + (trailingSlash ? '/' : ''), parent, job);
+  }
+  else {
     resolved = resolvePackage(specifier, parent, job);
+  }
   if (resolved.startsWith('node:')) return resolved;
   return job.realpath(resolved, parent);
 };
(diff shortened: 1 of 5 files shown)


src/resolve-dependency.js
```diff
@@ -18,8 +21,8 @@ function resolvePath (path, parent, job) {
 }
 
 function resolveFile (path, parent, job) {
+  if (path[path.length - 1] === '/') return;
```
styfle, line 24: I noticed you're using `/` here but `sep` below. Why is that?

guybedford: In this path, it was already normalized by the line number 9 of this file to always be `/`.

src/resolve-dependency.js
```diff
@@ -5,10 +5,13 @@ const { isAbsolute, resolve, sep } = require('path');
 // (package.json files are emitted as they are hit)
 module.exports = function resolveDependency (specifier, parent, job) {
   let resolved;
-  if (isAbsolute(specifier) || specifier === '.' || specifier === '..' || specifier.startsWith('./') || specifier.startsWith('../'))
-    resolved = resolvePath(resolve(parent, '..', specifier), parent, job);
-  else
+  if (isAbsolute(specifier) || specifier === '.' || specifier === '..' || specifier.startsWith('./') || specifier.startsWith('../')) {
+    const trailingSlash = specifier[specifier.length - 1] === '/' || specifier[specifier.lenggth - 1] === '\\';
```
styfle, line 9:
```suggestion
    const trailingSlash = specifier[specifier.length - 1] === '/' || specifier[specifier.length - 1] === '\\';
```

guybedford: Just saw that too, thanks for spotting.

src/resolve-dependency.js
styfle: So should we update line 27 and 28 to use `/` instead?

Also what about line 37?

guybedford: 27 and 28 run against the output of realpath which is then sep-based.

For inputs into file functions, `/` and `\\` can be used interchangeably.

It's worth noting too that `require('.\\asdf')` isn't even supported in Node.js itself, so users using backslashes is already a very rare with spotty support and shouldn't be done.

guybedford:
Test failure here seems like a flake?

src/resolve-dependency.js
styfle: Also, why change from `path.endsWith` to  `path[path.length - 1] ===`? I think the former reads better.

```suggestion
  if (path.endsWith('/')) return;
```

src/resolve-dependency.js
```diff
@@ -5,10 +5,13 @@ const { isAbsolute, resolve, sep } = require('path');
 // (package.json files are emitted as they are hit)
 module.exports = function resolveDependency (specifier, parent, job) {
   let resolved;
-  if (isAbsolute(specifier) || specifier === '.' || specifier === '..' || specifier.startsWith('./') || specifier.startsWith('../'))
-    resolved = resolvePath(resolve(parent, '..', specifier), parent, job);
-  else
+  if (isAbsolute(specifier) || specifier === '.' || specifier === '..' || specifier.startsWith('./') || specifier.startsWith('../')) {
+    const trailingSlash = specifier[specifier.length - 1] === '/' || specifier[specifier.length - 1] === '\\';
```
styfle, line 9:
```suggestion
    const trailingSlash = specifier.endsWith('/') || specifier.endsWith('\\');
```

Also, I don't see a unit test for the backslash (`\\`) use case?

guybedford: I double-checked in Node.js and it doesn't support a tailing `\`... so I've removed this test entirely.

src/resolve-dependency.js
```diff
@@ -29,6 +32,7 @@ function resolveFile (path, parent, job) {
 }
 
 function resolveDir (path, parent, job) {
+  if (path[path.length - 1] === '/') path = path.slice(0, -1);
```
styfle, line 35:
```suggestion
  if (path.endsWith('/')) path = path.slice(0, -1);
```

lukebennett:
This issue is currently blocking me from deploying a service to Vercel - any idea of how long it's likely to be until this fix trickles down to @vercel/node?

PR #129, opened 2020-06-26
Exports support, dynamic import from CJS support, package boundary emission

guybedford:
This PR implements three new features
* Support for the `exports` field as per Node.js, including conditional exports and package name self-resolution
* Support for emitting package.json files when they inform the module format interpretation of a file - this results in more package.json files being emitted, as can be seen in the unit test updates here
* Support for dynamic import in CJS files being traced

Exports support is by default disabled, and enabled by setting the `exports: true` option or providing a custom condition list.

For example, if building for Node.js 10, you'd want the exports support to be disabled since the `"main"` resolution would still apply - which would not be traced if enabling exports (when exports overrides it).


src/node-file-trace.js
```diff
@@ -73,6 +74,9 @@ class Job {
     }
     this.base = base;
     this.cwd = resolve(processCwd || base);
+    if (!Array.isArray(exports) && typeof exports !== 'boolean')
+      exports = false;
+    this.exports = exports === true ? ['node'] : exports;
```
styfle, line 79: I thought `exports: true` would mean all exports would get traced?

src/node-file-trace.js
```diff
@@ -224,6 +228,14 @@ class Job {
     return true;
   }
 
+  getPjsonBoundary (path) {
+    const root = path.slice(0, path.indexOf(sep)) || path.sep;
```
styfle, line 232: I would expect `root` to always be `path.sep`. It seems like this implementation will skip the system root directory.

Also, should this account for the `base` option? Maybe that should be considered the `root`?

src/resolve-dependency.js
```diff
@@ -56,18 +49,110 @@ function notFound (specifier, parent) {
...
+    const target = getExportsTarget(exports[subpath], job.exports, cjsResolve);
+    if (typeof target === 'string' && target.startsWith('./'))
+      return pkgPath + target.slice(1);
+  }
+  for (const match of Object.keys(exports)) {
+    if (!match.endsWith('/'))
+      continue;
+    if (subpath.startsWith(match)) {
+      const target = getExportsTarget(exports[match], job.exports, cjsResolve);
+      if (typeof target === 'string' && target[target.length - 1] === '/' && target.startsWith('./'))
```
styfle, line 108:
```suggestion
      if (typeof target === 'string' && target.endsWith('/') && target.startsWith('./'))
```

src/node-file-trace.js
guybedford: We only need to trace the resolution for the target environment. Eg no need to trace a "browser" or "react-native" build if we aren't building for that platform.

src/node-file-trace.js
guybedford: This is the boundary algorithm from https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_resolver_algorithm READ_PACKAGE_SCOPE.

The root should be `C:\\` on windows and `/` on posix. If I got the logic wrong this might be the cause of the loop. What is weird is I'm testing on WSL Ubuntu here so would have thought I'd seen that.

guybedford: Ah this likely was the cause - I wasn't running the ecmascript modules tests locally which exactly hit this root scenario. Test should be passing now.

src/node-file-trace.js
styfle: If that’s the case, the a better default is [node] instead of true and then the value can always be an Array.

styfle: This package is called node-file-trace so we should assume node unless the user opts out

guybedford: The problem is turning it on by default will break Node.js 10 support.

guybedford: Is it possible for the consumer-side to know the Node.js version? If not an alternative would be to support both Node.js versions without and without "exports" at the cost of extra tracing.

guybedford:
@styfle I've implemented an alternative that separates `"exports"` from `"exportsOnly"` as a separate feature.

The default behaviour now will trace both exports and the main.

Setting exports allows changing the list of exports conditions that are traced.

Then setting `exportsOnly: true` is a mode for Node.js 12+ only which will no longer trace the legacy main.

guybedford:
The remaining windows failures here are pretty odd - I can't replicate them on my Windows machine running the same Node.js version...

guybedford:
Finally traced the issue... this should be good to merge.

styfle:
@guybedford This doesn't work with `es-get-iterator@1.1.0`.

For example:

```
mkdir example && cd example
yarn init -y && yarn add es-get-iterator@1.1.0
echo 'require('es-get-iterator')' > start.js
nft print start.js
```

I would expect `node_modules/es-get-iterator/index.js` to be included but it is not.

Can we add an integration test for `es-get-iterator`?

guybedford:
Nicely spotted, ok I've added the integration test and also included a unit test for the array fallback case.

package.json
```diff
@@ -26,6 +26,7 @@
     "acorn-numeric-separator": "^0.3.0",
     "acorn-static-class-features": "^0.2.1",
     "bindings": "^1.4.0",
+    "es-get-iterator": "^1.1.0",
```
styfle, line 29: Should be `devDependencies` 🙂

src/resolve-dependency.js
```diff
@@ -78,8 +78,8 @@ function getExportsTarget (exports, conditions, cjsResolve) {
   }
   else if (Array.isArray(exports)) {
     for (const item of exports) {
-      const target = getExportsTarget(item);
-      if (target !== undefined)
+      const target = getExportsTarget(item, conditions, cjsResolve);
+      if (target !== undefined && target !== null && target.startsWith('./'))
```
styfle, line 82:
```suggestion
      if (typeof target === 'string' && target.startsWith('./'))
```

src/resolve-dependency.js
```diff
@@ -90,11 +90,14 @@ function getExportsTarget (exports, conditions, cjsResolve) {
           condition === 'import' && !cjsResolve ||
           conditions.includes(condition)) {
         const target = getExportsTarget(exports[condition], conditions, cjsResolve);
-        if (target !== undefined)
+        if (target !== undefined && (target === null || target.startsWith('./')))
```
styfle, line 93: Is it expected that this `if` statement does not match the one on line 82 above?

readme.md
```diff
@@ -57,6 +57,34 @@ const { fileList } = await nodeFileTrace(files, {
...
+
+This can be disabled with the `exportsOnly` option:
+
+```js
+const { fileList } = await nodeFileTrace(files, {
+  exportsOnly: true
+});
+```
+
+Any package with `"exports"` will then only have its exports traced, and the main will not be included at all.
```
styfle, line 86:
```suggestion
Any package with `"exports"` will then only have its exports traced, and the main will not be included at all. This can reduce the output size when targeting [Node.js 12.17.0)(https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#12.17.0) or newer.
```

package.json
guybedford: We're not currently pinning _any_ versions in this repo! I forget that's the preferred pattern, let's get a separate PR up for that.

src/resolve-dependency.js
guybedford: Ah, yes this is better :)

guybedford: Actually sorry you're completely right - it's actually incorrect that this is different to the below, pushing up a fix.

src/resolve-dependency.js
guybedford: Yes they should be the same - fixed.

styfle: @guybedford They are a little different still. Why use `continue` with one and not the other?

guybedford: Right, they are written differently but do the same thing. So the definition of the array fallback is that it will continue on invalid items, which is supposed to be different from the object conditions form. But I just realise now we're not actually doing that either! Done.

Note the extreme exports edge cases get pretty complex but aren't needed in 99% of cases... so we are just rattling out the `"exports": [{}, { "asdf": null }, null]` kinds of odd behaviours here.

styfle: @guybedford I don't see a test for the array of `exports` like you mentioned above? They are all object `exports` as far as I see. Can we add a test for array?

guybedford: Yes there is a test at https://github.com/vercel/node-file-trace/pull/129/files#diff-fb129c0544238468b2fbaf801b451a7f for the fallback support.

test/unit/yarn-workspaces/output.js
```diff
@@ -1,6 +1,8 @@
 [
+  "package.json",
   "test/unit/yarn-workspaces/input.js",
   "test/unit/yarn-workspaces/node_modules/x",
+  "test/unit/yarn-workspaces/node_modules/x/package.json",
```
styfle, line 5: @guybedford I think this file was incorrectly emitted since this line is a symlink pointing to the last line in the output here. See PR #135 for the fix.

PR #133, opened 2020-07-08
Change source code to TypeScript

styfle:
I found a few bugs when changing the source code to TypeScript.

I'll need some feedback from @guybedford to decide how to proceed in some cases, see comments below.


src/analyze.ts
```diff
@@ -399,17 +409,17 @@ module.exports = async function (id, code, job) {
...
     if (!staticChildNode) throw new Error('Internal error: No staticChildNode for backtrack.');
     const curStaticValue = computePureStaticValue(parent, true);
     if (curStaticValue) {
       if ('value' in curStaticValue && typeof curStaticValue.value !== 'symbol' ||
-          typeof curStaticValue.then !== 'symbol' && typeof curStaticValue.else !== 'symbol') {
+          'then' in curStaticValue && typeof curStaticValue.then !== 'symbol' && typeof curStaticValue.else !== 'symbol') {
         staticChildValue = curStaticValue;
         staticChildNode = parent;
-        if (self.skip) self.skip();
+        if (self && self.skip) self.skip();
```
styfle, line 422: I found that `self` was sometimes undefined. Is that expected?

src/resolve-dependency.ts
```diff
@@ -1,28 +1,34 @@
...
   }
-  if (typeof resolved === 'string' && resolved.startsWith('node:')) return resolved;
-  if (Array.isArray(resolved))
+  if (typeof resolved === 'string' && !resolved.startsWith('node:')) {
+    return job.realpath(resolved, parent);
+  } else if (Array.isArray(resolved)) {
     return resolved.map(resolved => job.realpath(resolved, parent));
-  return job.realpath(resolved, parent);
+  }
+  return resolved;
```
styfle, line 20: Is it expected that `resolved` can be undefined here?

src/types.ts
```diff
@@ -29,6 +29,8 @@ interface Stats {
 export interface NodeFileTraceOptions {
   base?: string;
   processCwd?: string;
+  exports?: string[];
+  exportsOnly?: boolean;
```
styfle, line 33: Added missing types here

src/types.ts
```diff
@@ -60,7 +62,15 @@ export interface NodeFileTraceResult {
...
 }
 
-export default function NodeFileTrace(
-  files: string[],
-  opts: NodeFileTraceOptions
-): Promise<NodeFileTraceResult>;
+export interface SingleValue {
+  value: string | string[] | undefined;
+  wildcards?: string[];
+}
```
styfle, line 68: I'm not sure if the `wildcards` option should be part of `SingleValue` or if I should make a new `WildcardValue` type?

src/utils/get-package-base.ts
```diff
@@ -22,6 +24,5 @@ module.exports.getPackageName = function (id) {
       return pkgNameMatch[0].replace(/\\/g,  '/');
     }
   }
+  return undefined;
```
styfle, line 27: There was an implicit return here so do we expect this to be undefined in some cases?

src/analyze.ts
```diff
@@ -399,17 +409,17 @@ module.exports = async function (id, code, job) {
   handleWrappers(ast);
   ({ ast = ast, scope = scope } = handleSpecialCases({ id, ast, scope, emitAsset: path => assets.add(path), emitAssetDirectory, job }) || {});
 
-  function backtrack (self, parent) {
+  function backtrack (self: WalkerContext, parent: Node) {
```
styfle, line 412: I found that `parent` can be undefined according to the typescript declarations in `walk`.

```js
  walk(ast, {
    enter (node, parent) {
     // ...
     }
```

Did you know this? How should we handle that case for `backtrack()` and `isIdentifierRead()`?

src/utils/get-package-base.ts
```diff
@@ -10,9 +11,10 @@ module.exports = function (id) {
     if (pkgNameMatch)
       return id.substr(0, pkgIndex + 13 + pkgNameMatch[0].length);
   }
-};
+  return undefined;
```
styfle, line 14: There was an implicit return here so do we expect this to be undefined in some cases?

src/analyze.ts
```diff
@@ -780,13 +789,14 @@ module.exports = async function (id, code, job) {
   }
 
   function emitStaticChildAsset () {
```
styfle, line 791: Should `emitStaticChildAsset()` accept any parameters? I see a couple usages like `emitStaticChildAsset(staticBindingsInstance)` and `emitStaticChildAsset(path)` that seem to indicate it should accept a parameter?

src/analyze.ts
guybedford: `parent` can only be undefined when `node` is a `Program`, and we never call `backtrack` on a `Program` node so this can't happen. Feel free to just add a forced typing as necessary here.

src/analyze.ts
guybedford: The WalkerContext should always be defined. Can you explain how it is ever coming up as undefined?

src/analyze.ts
guybedford: This is left over from ncc here - https://github.com/vercel/webpack-asset-relocator-loader/blob/master/src/asset-relocator.js#L1210.

These paths can all be removed as we don't need to manipulate output in this project so yes any argument computation code paths can be deleted as well.

src/resolve-dependency.ts
guybedford: `resolved` will never be undefined here, and the typing should be able to capture that?

src/types.ts
guybedford: Thanks!

src/utils/get-package-base.ts
guybedford: Yes, the undefined result is used for the "invalid" case.

src/utils/get-package-base.ts
guybedford: Yes this is correct.

src/types.ts
guybedford: This does seem simpler to me... maybe just call it `StaticValue` or something like that?

src/analyze.ts
```diff
@@ -381,32 +392,31 @@ module.exports = async function (id, code, job) {
...
         (isImport ? imports : deps).add(computed.else);
     }
   }
 
   let scope = attachScopes(ast, 'scope');
   handleWrappers(ast);
-  ({ ast = ast, scope = scope } = handleSpecialCases({ id, ast, scope, emitAsset: path => assets.add(path), emitAssetDirectory, job }) || {});
-
-  function backtrack (self, parent) {
+  handleSpecialCases({ id, ast, emitAsset: path => assets.add(path), emitAssetDirectory, job });
```
styfle, line 411: Is it expected that handleSpecialCases() doesn't return anything? The previous code made it seem like it was supposed to return `ast` and `scope` but `handleSpecialCases` does not have a return statement.

src/analyze.ts
styfle: Ok, I updated the calls to `emitStaticChildAsset()` so they no longer use any parameters.

src/resolve-dependency.ts
styfle: There's two calls above, one to `resolvePath()` and another to `resolvePackage()` and both can return undefined. What makes you think that `resolved` will never be `undefined` here?

src/resolve-dependency.ts
```diff
@@ -1,39 +1,46 @@
...
+function resolveFile (path: string, parent: string, job: Job): string | undefined {
+  if (path.endsWith('/')) return undefined;
   path = job.realpath(path, parent);
   if (job.isFile(path)) return path;
   if (job.ts && path.startsWith(job.base) && path.substr(job.base.length).indexOf(sep + 'node_modules' + sep) === -1 && job.isFile(path + '.ts')) return path + '.ts';
   if (job.ts && path.startsWith(job.base) && path.substr(job.base.length).indexOf(sep + 'node_modules' + sep) === -1 && job.isFile(path + '.tsx')) return path + '.tsx';
   if (job.isFile(path + '.js')) return path + '.js';
   if (job.isFile(path + '.json')) return path + '.json';
   if (job.isFile(path + '.node')) return path + '.node';
+  return undefined;
```
styfle, line 40: Should `resolveFile()` return undefined or should it throw NotFoundError?

src/resolve-dependency.ts
```diff
@@ -47,35 +54,40 @@ function resolveDir (path, parent, job) {
...
-  const e = new Error("Cannot find module '" + specifier + "' loaded from " + parent);
-  e.code = 'MODULE_NOT_FOUND';
-  throw e;
+class NotFoundError extends Error {
+  public code: string;
+  constructor(specifier: string, parent: string) {
+    super("Cannot find module '" + specifier + "' loaded from " + parent);
+    this.code = 'MODULE_NOT_FOUND';
+  }
 }
```
styfle, line 63: I changed the `notFound()` function to a class because typescript doesn't keep track if a function throws or not so we need to throw this custom error where appropriate.

src/resolve-dependency.ts
```diff
@@ -95,15 +107,19 @@ function getExportsTarget (exports, conditions, cjsResolve) {
       }
     }
   }
-  else if (exports === null) {
-    return exports;
-  }
+
+  return null;
```
styfle, line 111: Is it okay to return null here or should it be undefined?

src/resolve-dependency.ts
```diff
@@ -95,15 +107,19 @@ function getExportsTarget (exports, conditions, cjsResolve) {
...
 }
 
-function resolveExportsTarget (pkgPath, exports, subpath, job, cjsResolve) {
-  if (typeof exports === 'string' ||
-      typeof exports === 'object' && !Array.isArray(exports) && Object.keys(exports).length && Object.keys(exports)[0][0] !== '.')
-    exports = { '.' : exports };
+function resolveExportsTarget (pkgPath: string, exp: string | { [key: string]: string }, subpath: string, job: Job, cjsResolve: boolean): string | undefined {
```
styfle, line 114: I noticed that `getExportsTarget()` handles the `exports: null` case however this `resolveExportsTarget()` function does not. Should we also handle `exports: null` here?

src/analyze.ts
styfle: I have no idea how I broke this. I must have changed `this` somehow but I can't seem to figure out where 🤔 

You can see the `asset-conditional` test fail without the check for `self` existence:

https://github.com/vercel/node-file-trace/pull/133/checks?check_run_id=862034673#step:8:111

Any ideas?

src/analyze.ts
styfle: Just to be clear, the only time `parent` is undefined is when `node.type === 'Program'`, correct?

src/analyze.ts
styfle: Could this be a bug in the old version of `estree-walker` we are using?

I noticed that `enter` has the correct context but not `leave` which is where it is failing.

When I looked at the source code of `estree-walker` I see `enter.call(context, node, parent, prop, index);` and then `leave(node, parent, prop, index);` which makes me think its a bug in `estree-walker`.

Its still not clear to me why we didn't see this test failing before though 🤔

styfle: I compared the stack trace: the left side is `master` and the right side is my PR.

diff

These are both using the same version of `estree-walker` 🤔

src/analyze.ts
```diff
@@ -708,7 +719,9 @@ module.exports = async function (id, code, job) {
     },
     leave (node, parent) {
       if (node.scope) {
-        scope = scope.parent;
+        if (scope.parent) {
+          scope = scope.parent;
+        }
```
styfle, line 724: Is it safe to assume that we don't want to assign the scope to `undefined` when the `scope.parent` is undefined?

src/analyze.ts
guybedford: Yeah it's fine to remove the return paths.

src/resolve-dependency.ts
guybedford: Both `resolvePath` and `resolvePackage` should never return `undefined`.

src/resolve-dependency.ts
guybedford: It should return undefined.

src/resolve-dependency.ts
guybedford: An alternative might be make the return value `never` I believe.

src/resolve-dependency.ts
guybedford: It needs to be `undefined` not `null`. They have different meanings for this function.

src/resolve-dependency.ts
guybedford: Yes on 117 we should probably have a null check too for consistency.

src/analyze.ts
guybedford: Yes exactly.

src/analyze.ts
guybedford: So just like `enter`, the only time `parent` is undefined is when `node` is `Program`. So yes this would assign `undefined` to scope when leaving at the very end of the run, but no other code checking `scope` would run after that.

So it's fine to either include this check or to enforce that scope is always defined since the case where it is not defined is effectively non-executable.

src/analyze.ts
guybedford: Right, as you say the `leave` function does not have a skip option so that seems to be what is happening. So perhaps previously the `this` value was the global default and now it isn't or something like that. So, having the explicit context check for the enter case only makes sense as you have provided here.

Note that we can update the check though to `if (self)` instead of `if (self.skip)` since `self.skip` will be defined whenever `self` is defined, so we're just updating the entire check for the new `this` scoping.

src/resolve-dependency.ts
styfle: @guybedford But `resolvePath()` calls `resolveFile()` and [you mentioned](https://github.com/vercel/node-file-trace/pull/133#discussion_r453301867) that `resolveFile()` can return undefined.

Therefore, `resolvePath()` can return `undefined` and this `undefined` case is not handled in `resolveDependency()` so `resolveDependency()` can returned `undefined` as well.

Perhaps typescript is revealing a case we didn't think about and it should be handled?

src/resolve-dependency.ts
styfle: What should happen for `null` on line 117?

- `return null`?
- assign `exports: { '.' : null }`?
- assign `exports: null`?
- something else?

src/resolve-dependency.ts
styfle: Thanks, updated to `return undefined` 👍

src/analyze.ts
styfle: Thanks, I updated to `backtrack()` to accept the optional context and use `if (context) context.skip()`

src/resolve-dependency.ts
guybedford: Actually `exports` can never be `null` or `undefined` here as it is guarded in all calls.

src/resolve-dependency.ts
guybedford: `resolvePath` will always explicitly throw an error for no resolution, and can't return `undefined`, while `resolveFile` can.

styfle: @guybedford What about `resolvePackage()`? It returns `resolveFile()` which could be undefined.

That still means that `resolveDependency()` can return undefined as well.

src/resolve-dependency.ts
```diff
@@ -183,5 +200,5 @@ function resolvePackage (name, parent, job, cjsResolve) {
       return resolveFile(pathTarget, parent, job) || resolveDir(pathTarget, parent, job);
```
guybedford, line 200: This path should have `|| throwNotFound` on the fallback.

src/resolve-dependency.ts
guybedford: No, `resolvePackage` should always return a result or throw. If there are any paths that don't do that that is a bug - I've just commented on one possible case.

src/resolve-dependency.ts
styfle: Good point, I added a `PkgCfg` and `Exports` type so we can confirm all cases are covered, thanks!

src/resolve-dependency.ts
styfle: Ah ha, this is the missing piece! I pushed a change which makes sure `resolveDependency(): string | string[]` so we never return undefined anymore ☺️

src/resolve-dependency.ts
styfle: Looks good now after https://github.com/vercel/node-file-trace/pull/133#discussion_r453679297 👍

PR #135, opened 2020-07-15
Fix package.json emit for symlinked yarn workspaces

styfle:
PR #129 emits `package.json` files but it should not be emit if package.json is inside a symlinked directory. This is a common pattern for yarn workspaces.

This PR fixes yarn workspaces and adds a new test for yarn workspaces using ESM imports/exports.


test/unit/yarn-workspaces/output.js
```diff
@@ -2,7 +2,6 @@
   "package.json",
   "test/unit/yarn-workspaces/input.js",
   "test/unit/yarn-workspaces/node_modules/x",
-  "test/unit/yarn-workspaces/node_modules/x/package.json",
```
styfle, line 5: I believe this was incorrectly added in https://github.com/vercel/node-file-trace/pull/129/files#r455316015

src/resolve-dependency.ts
```diff
@@ -50,7 +50,7 @@ function resolveDir (path: string, parent: string, job: Job) {
   if (pkgCfg && typeof pkgCfg.main === 'string') {
     const resolved = resolveFile(resolve(path, pkgCfg.main), parent, job) || resolveFile(resolve(path, pkgCfg.main, 'index'), parent, job);
     if (resolved) {
-      job.emitFile(path + sep + 'package.json', 'resolve', parent);
+      job.emitFile(job.realpath(path, parent) + sep + 'package.json', 'resolve', parent);
```
guybedford, line 53: Maybe we should solve this a little more generally and instead have `emitFile` do realpathing and also handle internal emission of symlinks?

guybedford:
Nicely worked through.... it may be nicer to do a more general refactoring here I think.

Let me know your thoughts on that.

src/resolve-dependency.ts
styfle: I tried changing the first line of `emitFile()` to be `path = this.realpath(path, parent)` but that broke almost all the tests.

styfle:
@guybedford I'm surprised to see Windows fail from this change. Any idea what that is? 🤔 

https://github.com/vercel/node-file-trace/pull/135/checks?check_run_id=874918207#step:9:164

src/resolve-dependency.ts
guybedford: Yes that is why it needs to be done alongside a symlink emission mechanism.

guybedford: Hmm realpath is supposed to be handling symlink emission already actually. It needs some deeper investigation it seems.

styfle: It looks like `realpath()` calls `emitFile()` already https://github.com/vercel/node-file-trace/blob/265dc2ad16b409aa5a6bf7bd05e11b5b0eeac026/src/node-file-trace.ts#L227

styfle: It doesn't seem correct to call `realpath()` from inside `emitFile()` so I'm going to have to leave it as is.

The only other place I see that might need it is here: https://github.com/vercel/node-file-trace/blob/dcf5538c0d0b4dd508b2cef0cef414197f0ea9d1/src/resolve-dependency.ts#L168

and perhaps here:

https://github.com/vercel/node-file-trace/blob/dcf5538c0d0b4dd508b2cef0cef414197f0ea9d1/src/resolve-dependency.ts#L190

test/unit/yarn-workspaces/output.js
guybedford: Agreed.

src/resolve-dependency.ts
guybedford: I'm still not 100% sure I follow the issue here, so it would help if you could walk through it with me.

The bug is that symlinked package.json files are being emitted? Why is this a bug? It seems correct to me that symlinks should be emitted when they exist in the path that is being imported to define the package boundary. I'm not clear on why emitting them is calling a failure in the first place at all either?

The fix seems to be to emit the original package.json file, but not emit the symlink itself. That is the creation of the symlink itself seems to be the bug right? Why is that? What about having a symlinked package.json file is causing a failure in workspaces?

styfle: The failure is happening in `@vercel/next` https://github.com/vercel/vercel/pull/4814

Previously, we would only emit the symlinked directory but nothing inside of it. In this case `node_modules/@example/common` is a symlink pointing to `node_modules/packages/common` so we emit that as usual.

The problem is that we were also emitting `node_modules/@example/common/package.json` which is not a symlink and doesn't exist.

For example

- `fs.readlinkSync('node_modules/@example/common')` prints `../../packages/common`
- `fs.readlinkSync('node_modules/@example/common/package.json')` throws EINVAL

Does that make sense?

guybedford: Yes that makes sense - so the problem is we are treating file symlinks and junctions the same. It sounds like the fix should then be to support proper emission of junctions and symlinks separately?

guybedford, line 53:
```suggestion
      job.emitFile(job.realpath(path + sep + 'package.json', parent), 'resolve', parent);
```

This will handle package.json symlinks properly.

styfle:
I like the idea of emitFile() calling realpath() just in case we missed something, lets see if the tests pass and then ✅

PR #136, opened 2020-07-16
Fix ts exports and code coverage

styfle:
- Created `index.ts` file as the main entrypoint so we can selectively export types, etc
- Fixed code coverage which was not pattern matching any files
- Added coverage threshold so CI fails if we ever drop below our current coverage

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,8 +3,8 @@
   "version": "0.7.0",
   "repository": "vercel/node-file-trace",
   "license": "MIT",
-  "main": "./out/node-file-trace.js",
-  "types": "./out/node-file-trace.d.ts",
+  "main": "./out/index.js",
+  "types": "./out/index.d.ts",
   "bin": {
     "nft": "./out/cli.js"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -3,18 +3,17 @@
...
     "nft": "./out/cli.js"
   },
   "scripts": {
     "build": "tsc",
     "prepublishOnly": "tsc && rm out/tsconfig.tsbuildinfo",
     "test": "jest --verbose",
-    "test-verbose": "jest --verbose --coverage --globals \"{\\\"coverage\\\":true}\"",
-    "codecov": "codecov",
-    "test-coverage": "jest --verbose --coverage --globals \"{\\\"coverage\\\":true}\" && codecov"
+    "test-verbose": "tsc --sourceMap && jest --verbose --coverage --globals \"{\\\"coverage\\\":true}\"",
```
TooTallNate, line 15: I think this should work too:

```suggestion
    "test-verbose": "tsc --sourceMap && jest --verbose --coverage --globals '{\"coverage\":true}'",
```

PR #148, opened 2020-07-30
Ensure webpack wrapper statement existence checks

guybedford:
As a follow-up to https://github.com/vercel/node-file-trace/pull/147 I just realised these statement checks can in theory cause undefined errors (like some previous bug cases).

This updates those checks to always ensure existence so they don't fail.


guybedford:
@styfle I actually just looked into this and it turns out that we have already guarded this case fine! Better to check these things though at least. Closing.

PR #152, opened 2020-08-19
Resolve yarn pnp dependencies

sergioramos:
I'm not sure how to test this, but let me describe the problem and solution that I'm proposing.

---

When using nft on a codebase with Yarn Berry _AND_ pnp active, the normal resolution of modules won't work as it will try to go though `node_modules` folders that just don't exist.

This will give it a correct resolution but it won't be able to read the file, unless run with `pnpify`. I feel like that shouldn't be the responsibility of nft, especially given that it exposes ways to hook into `readFile`.

The way I'm planning to use this is though [serverless-plugin-tree-shake](https://github.com/sergioramos/serverless-plugin-tree-shake). After having the file list, when writing it into a zip I intend to use `@yarnpkg/pnpify`'s `buildNodeModulesTree` to map the virtual files and write into an usable tree under `node_modules`.


src/resolve-dependency.ts
```diff
@@ -176,6 +181,12 @@ function resolvePackage (name: string, parent: string, job: Job, cjsResolve: boo
     }
   }
 
+  if (resolveRequest) {
```
styfle, line 184:
```suggestion
  if (typeof resolveRequest === 'function') {
```

sergioramos: fixed, thank you 👍

src/resolve-dependency.ts
```diff
@@ -176,6 +183,12 @@ function resolvePackage (name: string, parent: string, job: Job, cjsResolve: boo
     }
   }
 
+  if (typeof resolveRequest === 'function') {
+    try {
+      return resolveRequest(name, parent);
```
merceyz, line 188: Can't this function just use `require.resolve(name, { paths: [parent] });` to get the entry point for each `name`? That would make it not be PnP specific

merceyz (commented):
The original implementation should use `require.resolve` to get the entrypoint for a package instead of relying on `node_modules/name` to exist.

> I think PnP is a terrible idea

Care to elaborate on that belief?

test/integration.test.js
```diff
@@ -54,3 +56,28 @@ for (const integrationTest of readdirSync(integrationDir)) {
...
+
+  // it needs to run within that context because otherwise require('pnpapi') won't work
+  const { stdout } = await exec('yarn pnpify node trace.js', { cwd });
+
+  const { fileList = [] } = JSON.parse(stdout);
+
+  const tree = fileList.sort().map((pathname) => {
+    return pathname
+      .split(path.sep)
+      .map((part) => (path.extname(part) === '.zip' ? 'generated.zip' : part))
```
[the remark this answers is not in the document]
styfle: @guybedford Isn't that basically what its doing? How does `extname()` differ from `endsWith()`?

src/resolve-dependency.ts
sergioramos: that is a good idea, let me check that to see if it works @merceyz.

sergioramos:
@guybedford tbh I can't say I'm a fan of pnp either, but that's another "battle". And I also like the idea of being able to hook into the module resolution. I'm going to try to use module.resolve for this PR and try to create another one to add an hook to resolution.

sergioramos:
@guybedford @styfle updated the PR to use require.resolve and improved the test case to ensure multiple versions of the same package are resolved properly.

src/resolve-dependency.ts
```diff
@@ -176,6 +176,12 @@ function resolvePackage (name: string, parent: string, job: Job, cjsResolve: boo
     }
   }
 
+  if ('pnp' in process.versions) {
+    try {
+      return require.resolve(name, { paths: [parent] });
+    } catch (err) {}
+  }
+
```
merceyz, line 184: This is still PnP specific, this entire function should just use `require.resolve` to get the entry point (what is currently hardcoded to `<packageParent>/node_modules/name`) then continue its logic as is

sergioramos: yes, but I don't know why the current logic doesn't do that, there could be a reason for the current resolve logic to be the way it is. I didn't want to break anything 😉

guybedford:
Great to see you've been able to make a hook approach work. The concept I was suggesting for a resolver hook would be something more like:

```js
const { fileList } = await nodeFileTrace(files, {
  resolve: customResolverFunction
}
```

where the default `resolve` is the one in `resolve-dependency.js`.

In terms of the arguments against PnP, I forget the exact details of the argument now, but when I was thinking about it I remember the issue is that it relies on hooking things like fs functions etc in Node.js which can never be a comprehensive virtualization since there are edge cases to this process.

sergioramos:
@guybedford every time I've used pnp, the install speedups never paid off for the troubles and inconveniences that it caused. 

Yeah, I'm working on a PR for it: #153

sergioramos:
Converted this PR into a draft while #153 is active.

PR #153, opened 2020-08-25
Expose hook into the module resolution (`resolve`)

sergioramos:


src/types.ts
```diff
@@ -45,6 +45,7 @@ export interface NodeFileTraceOptions {
   readFile?: (path: string) => Buffer | string | null;
   stat?: (path: string) => Stats | null;
   readlink?: (path: string) => string | null;
+  resolve?: (id: string, parent: string) => Promise<string | string[]>;
```
styfle, line 48: Does this need to be async? The other options are synchronous and I thought your previous PR was synchronous too?

sergioramos: I was confused by the documentation that has `readFile` and `readLink` as promises 🤦

sergioramos:
@styfle tests passing now 😉

src/resolve-dependency.ts
```diff
@@ -9,13 +9,14 @@ export default function resolveDependency (specifier: string, parent: string, jo
...
-  else {
+  } else if (typeof job.resolve === 'function') {
+    resolved = job.resolve(specifier, parent);
+  } else {
     resolved = resolvePackage(specifier, parent, job, cjsResolve);
   }
 
   if (Array.isArray(resolved)) {
-    return resolved.map(resolved => job.realpath(resolved, parent));
+    return resolved.filter(Boolean).map(resolved => job.realpath(resolved, parent));
```
guybedford, line 19: Can you clarify why this is added here?

src/node-file-trace.ts
```diff
@@ -22,6 +22,8 @@ export async function nodeFileTrace(files: string[], opts: NodeFileTraceOptions
     job.stat = opts.stat;
   if (opts.readlink)
     job.readlink = opts.readlink;
+  if (opts.resolve)
+    job.resolve = opts.resolve;
```
guybedford, line 26: I hate to be too pedantic, but can we implement the model here as `job.resolve = opts.resolve || defaultResolve`, where `job.resolve` is the resolve function that is called in the first place?

test/fixtures/yarn-pnp/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "yarn-pnp-test",
```
guybedford, line 2: Do we still need this test here - can we not just have a simple resolve hook unit test rather now?

.github/workflows/ci.yml
```diff
@@ -39,7 +39,11 @@ jobs:
     - name: Run Tests
       env:
         BULL_REDIS_CONNECTION: ${{ secrets.BULL_REDIS_CONNECTION }}
-      run: yarn test-verbose
+      # we need to install deps on yarn-pnp _after_ we build nft because it has
+      # a virtual dependency on nft
+      run: |
+        cd test/fixtures/yarn-pnp && yarn || true; cd -
+        yarn test-verbose
```
styfle, line 46: I still think this part needs to be in `integration.test.js`

src/resolve-dependency.ts
sergioramos: because now we're dealing with outside input (that might not be typed). this is just a small assurance that the array has truthy things in  it before using them in `realpath()`

styfle: Seems like that should be validated earlier?

test/fixtures/yarn-pnp/package.json
sergioramos: removed 👍

src/node-file-trace.ts
sergioramos: implemented that suggestion 👍

sergioramos:
Implemented the suggestions and got the tests passing again 😉

src/node-file-trace.ts
```diff
@@ -65,7 +66,8 @@ export class Job {
   public processed: Set<string>;
   public warnings: Set<Error>;
   public reasons: NodeFileTraceReasons = Object.create(null);
-  
+  resolve!: (id: string, parent: string, job: Job, cjsResolve: boolean) => string | string[];
```
styfle, line 69: Can we be explicit if this is public/private?

Also why `!` here?

test/unit.test.js
```diff
@@ -1,9 +1,22 @@
 const fs = require('fs');
-const { join } = require('path');
+const { join, isAbsolute, resolve } = require('path');
 const { nodeFileTrace } = require('../out/node-file-trace');
 
 global._unit = true;
 
+const nodeBuiltins = [...require('repl')._builtinLibs, 'constants', 'module', 'timers', 'console', '_stream_writable', '_stream_readable', '_stream_duplex', 'process', 'sys'];
+const resolveDependency = (id, parent, ...rest) => {
```
styfle, line 8: This seems too complex for the test. Maybe just do something like:

```js
const resolveDependency = (id, parent) => `custom-resolution-${id}-${parent}`;
```

test/unit/resolve-hook/input.js
```diff
@@ -0,0 +1,2 @@
+require('./dep');
+require('when');
```
styfle, line 2: I don't think this test should depend on `when`, its not necessary to test the hook.

src/node-file-trace.ts
sergioramos: fair 👍

test/unit.test.js
sergioramos: fair 👍

test/unit/resolve-hook/input.js
sergioramos: https://github.com/vercel/nft/pull/153#discussion_r485210888

src/node-file-trace.ts
sergioramos: good call, made me revisit the implementation 👍

styfle (approved):
Great work, thanks!

PR #155, opened 2020-09-04
Fix symlinks when `base: "/"`

styfle:
We found a bug when `base: "/"` which was causing symlinks to not resolve properly.

Using test 40 from `vercel/vercel` repo:

### Before

```
[
  'Users/styfle/Code/foo/40-yarn-workspaces/api/index.js',
  'Users/styfle/Code/foo/40-yarn-workspaces/api/package.json',
  'Users/styfle/Code/foo/40-yarn-workspaces/node_modules/@builders-test/lib',
  'Users/styfle/Code/foo/40-yarn-workspaces/node_modules/@builders-test/lib/index.js',
  'Users/styfle/Code/foo/40-yarn-workspaces/node_modules/@builders-test/lib/package.json'
]
```

### After

```
[
  'Users/styfle/Code/foo/40-yarn-workspaces/api/index.js',
  'Users/styfle/Code/foo/40-yarn-workspaces/api/package.json',
  'Users/styfle/Code/foo/40-yarn-workspaces/node_modules/@builders-test/lib',
  'Users/styfle/Code/foo/40-yarn-workspaces/packages/lib/index.js',
  'Users/styfle/Code/foo/40-yarn-workspaces/packages/lib/package.json'
]
```


TooTallNate (approved):

src/node-file-trace.ts
```diff
@@ -5,12 +5,14 @@ import analyze, { AnalyzeResult } from './analyze';
...
 import { sharedLibEmit } from './utils/sharedlib-emit';
+import { join } from 'path';
 
 const { gracefulify } = require('graceful-fs');
 gracefulify(fs);
 
 function inPath (path: string, parent: string) {
-  return path.startsWith(parent) && path[parent.length] === sep;
+  const pathWithSep = join(parent, sep);
+  return path.startsWith(pathWithSep) && path !== pathWithSep;
```
guybedford, line 15:
```suggestion
  return path.startsWith(parent) && path[parent.length] === sep && path.length > parent.length;
```

seems equivalent and a slightly smaller diff, but otherwise seems gd.

PR #156, opened 2020-09-04
Run all unit tests against root

styfle:
Follow up to #155 which caused regression with other tests.

Instead, we need to run all tests against the root as well as the cwd.


test/unit/yarn-workspaces-base-root/.gitignore
```diff
@@ -1,2 +0,0 @@
-# include node_modules to test the symlinks
-!node_modules
```
styfle, line 2: I'm deleting this test because it was copy/paste of `test/unit/yarn-workspaces`

Timer (approved):

styfle:
For posterity, I found that the [`phantomjs-prebuilt`](https://github.com/vercel/nft/runs/1072588817#step:9:366) test failed because [GitHub Actions](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md) has PhantomJS preinstalled.

guybedford:
Ah perfect, that explains what changed! In theory if the emission is skipped I think it should still work though if the logic isn't rewritten? Might be something worth looking into if not, but otherwise the workaround here seems gd.

Issue #159, opened 2020-11-19
nft doesn't pick up certain `path.join` statements


timsuchanek:
It seems like if the `path.join` statements are in a specific function scope, they're not picked up by nft.
Reproduction here: https://github.com/timsuchanek/prisma-vercel-next

styfle:
Shipping here: https://github.com/vercel/vercel/pull/5455

PR #160, opened 2020-11-20
Fix webpack dirname wrapper analysis

guybedford:
Webpack 5 includes a scoping wrapper for `__dirname` which interfered with the analysis on the `path` binding in the emission.

This fixes up that emit resolving https://github.com/vercel/nft/issues/159.


styfle (approved):
Thanks!

PR #167, opened 2021-02-04
fix: update CI to test Node.js 14

styfle:
Also drop Node.js 10 which is about to reach EOL


styfle:
@guybedford Thoughts on this one? Should work if we update all the dependencies, right?

guybedford:
Yeah I think it requires updating the related integration tests.

Issue #151, opened 2020-08-12
Upgrade acorn to 8.0.0 to add support for ES2021 (Node 14.x)


styfle:
Changelog: https://github.com/acornjs/acorn/blob/master/acorn/CHANGELOG.md#800-2020-08-12

We can also add support for Optional Chaining which has a test but is currently ignored https://github.com/vercel/node-file-trace/blob/c341820924ef0a124e29351332067de5dfc06cf3/test/ecmascript.test.js#L20

Node 14 supports Optional Chaining, etc: https://node.green

PR #177, opened 2021-03-21
fix: test optional chaining

styfle:
Related to #151 and #176


styfle:
@guybedford Any others we can start testing for?

guybedford:
As far as I can tell I think that's it actually.

guybedford (approved):

PR #179, opened 2021-03-22
feat: resolve "imports" and "exports" wildcards

guybedford:
This updates the resolver to the latest Node.js resolver handling per the modules stability with "imports" support and "exports" wildcards support.

In the API, the `"exports"` option has been renamed to `"conditions"` as it applies to both "imports" and "exports", but backwards compatibility is provided for the former exports option so this is non-breaking.

### DIFF
--- a/src/resolve-dependency.ts
+++ b/src/resolve-dependency.ts
@@ -145,7 +145,7 @@ function resolveExportsImports (pkgPath: string, obj: PackageTarget, subpath: st
     if (typeof target === 'string' && target.startsWith('./'))
       return pkgPath + target.slice(1);
   }
-  for (const match of Object.keys(matchObj).sort((a, b) => a.length > b.length ? -1 : 1)) {
+  for (const match of Object.keys(matchObj).sort((a, b) => b.length - a.length)) {
     if (match.endsWith('*') && subpath.startsWith(match.slice(0, -1))) {
       const target = getExportsTarget(matchObj[match], job.conditions, cjsResolve);
       if (typeof target === 'string' && target.startsWith('./'))
--- a/src/resolve-dependency.ts
+++ b/src/resolve-dependency.ts
@@ -10,6 +10,9 @@ export default function resolveDependency (specifier: string, parent: string, jo
     const trailingSlash = specifier.endsWith('/');
     resolved = resolvePath(resolve(parent, '..', specifier) + (trailingSlash ? '/' : ''), parent, job);
   }
+  else if (specifier[0] === '#') {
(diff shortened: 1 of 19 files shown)


src/resolve-dependency.ts
```diff
@@ -123,32 +127,65 @@ function getExportsTarget(exports: string | string[] | { [key: string]: string }
...
   
-  if (subpath in exports) {
-    const target = getExportsTarget(exports[subpath], job.exports, cjsResolve);
+  if (subpath in matchObj) {
+    const target = getExportsTarget(matchObj[subpath], job.conditions, cjsResolve);
     if (typeof target === 'string' && target.startsWith('./'))
       return pkgPath + target.slice(1);
   }
-  for (const match of Object.keys(exports)) {
+  for (const match of Object.keys(matchObj).sort((a, b) => a.length > b.length ? -1 : 1)) {
```
styfle, line 148:
```suggestion
  for (const match of Object.keys(matchObj).sort((a, b) => b.length - a.length)) {
```

Slightly shorter way to do the same

src/resolve-dependency.ts
```diff
@@ -123,32 +127,65 @@ function getExportsTarget(exports: string | string[] | { [key: string]: string }
...
   return undefined;
 }
 
+function packageImportsResolve (name: string, parent: string, job: Job, cjsResolve: boolean): string {
+  if (name !== '#' && !name.startsWith('#/') && job.conditions) {
+    const pjsonBoundary = job.getPjsonBoundary(parent);
+    if (pjsonBoundary) {
+      const pkgCfg = getPkgCfg(pjsonBoundary, job);
+      const { imports: pkgImports } = pkgCfg || {};
+      if (pkgCfg && pkgImports !== null && pkgImports !== undefined) {
```
styfle, line 171: I noticed this line (171) is checking the existence of `pkgCfg` but line 225 is not.

Is that intentional?

guybedford: Yeah it's a redundant check actually. Should be fine to leave.

test/unit/exports-wildcard/input.js
```diff
@@ -0,0 +1,2 @@
+import { y } from 'y/no';
```
styfle, line 1: Does this code even run? It looks like `no.js` exports `x`, not `y` so how can you import `x` here?

guybedford: The exports wildcard remapping maps it to "node.js".

styfle (approved):

PR #180, opened 2021-04-03
Support import.meta.url asset emission

guybedford:
This adds support for `import.meta.url` based asset emission in NFT.

The approach is to treat `import.meta.url` as an asset trigger, which then emits path references just like `__filename` and `__dirname`.

Support is included for `new URL('./path', import.meta.url)` patterns as well with global analysis of the `URL` object.

`pathToFileURL` and `fileURLToPath` work out without custom handling since they are effectively irrelevant in the asset emission approach used here.

Quite advanced patterns are supported, eg a test is included for the following case:

```js
import { readFileSync } from 'fs';
import { fileURLToPath } from 'url';

console.log(readFileSync(
  fileURLToPath(`${import.meta.url}/../${unknownCondition ? 'asset1.txt' : 'asset2.txt'}`)
).toString()); 
```

emitting both `asset1.txt` and `asset2.txt` as possible references.

### DIFF
--- a/src/analyze.ts
+++ b/src/analyze.ts
@@ -15,7 +15,7 @@ import resolve from './resolve-dependency.js';
 //@ts-ignore
 import nodeGypBuild from 'node-gyp-build';
 import { Job } from './node-file-trace';
-import { fileURLToPath, pathToFileURL } from 'url';
+import { fileURLToPath, pathToFileURL, URL } from 'url';
 
 // Note: these should be deprecated over time as they ship in Acorn core
 const acorn = Parser.extend(
@@ -464,8 +464,11 @@ export default async function analyze(id: string, code: string, job: Job): Promi
       // currently backtracking
       if (staticChildNode) return;
 
+      if (!parent)
+        return;
+
       if (node.type === 'Identifier') {
-        if (isIdentifierRead(node, parent!) && job.analysis.computeFileReferences) {
+        if (isIdentifierRead(node, parent) && job.analysis.computeFileReferences) {
           let binding;
           // detect asset leaf expression triggers (if not already)
           // __dirname,  __filename
@@ -473,7 +476,7 @@ export default async function analyze(id: string, code: string, job: Job): Promi
               binding && (typeof binding === 'function' || typeof binding === 'object') && binding[TRIGGER]) {
             staticChildValue = { value: typeof binding === 'string' ? binding : undefined };
             staticChildNode = node;
-            backtrack(parent!, this);
+            backtrack(parent, this);
           }
         }
       }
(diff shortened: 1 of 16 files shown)


src/analyze.ts
```diff
@@ -453,6 +472,12 @@ export default async function analyze(id: string, code: string, job: Job): Promi
           }
         }
       }
+      else if (job.analysis.computeFileReferences && node.type === 'MemberExpression' && node.object.type === 'MetaProperty' && node.object.meta.name === 'import' && node.object.property.name === 'meta' && (node.property.computed ? node.property.value : node.property.name) === 'url') {
+        // import.meta.url leaf trigger
+        staticChildValue = { value: importMetaUrl };
+        staticChildNode = node;
+        backtrack(parent!, this);
```
styfle, line 479: The `!` seems like an error.

We should probably have `&& parent` in the `if` condition above if the parent is possibly null or undefined.

src/analyze.ts
```diff
@@ -15,6 +15,7 @@ import resolve from './resolve-dependency.js';
 //@ts-ignore
 import nodeGypBuild from 'node-gyp-build';
 import { Job } from './node-file-trace';
+import { fileURLToPath } from 'url';
```
styfle, line 18: Maybe add `pathToFileURL` here to so both are imported on the same line instead of two separate lines?

jest.config.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
       branches: 80.5,
       functions: 95.28,
       lines: 85.87,
-      statements: -249
+      statements: -260
```
styfle, line 9: Can we add tests to avoid reducing test coverage?

src/analyze.ts
```diff
@@ -791,32 +816,36 @@ export default async function analyze(id: string, code: string, job: Job): Promi
     return true;
   }
 
+  function resolveAbsolutePathOrUrl (value: string | URL): string {
+    return value instanceof URL ? fileURLToPath(value as any) : value.startsWith('file:') ? fileURLToPath(new URL(value) as any) : path.resolve(value);
```
styfle, line 820: You can remove the `as any` usages here if you do `import { URL } from 'url'` above, but not sure if this would cause other problems.

tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "incremental": true,
-    "lib": ["esnext"],
+    "lib": ["esnext", "DOM"],
```
styfle, line 7: I think you can remove the `DOM` lib here by using `import { URL } from 'url'`

styfle (approved):
Great work, thanks!

PR #184, opened 2021-04-12
test: added test for playwright-core

mxschmitt:
This PR adds a test to demonstrate a current issue which was reported [here](https://github.com/microsoft/playwright/issues/5862).

TL;DR: The following file does not get added: `node_modules/playwright-core/browsers.json`

See here for more information and background: https://github.com/microsoft/playwright/issues/5862

The relevant part in the code is the `fs.readFileSync` which does not get detected. If you use `require`
 instead it works and it does get added.

The following line is the relevant one in Playwright, if it does get changed to `require` instead, it does pass the test. (change it in `node_modules/playwright-core/lib/utils/registry.js` for local testing)

https://github.com/microsoft/playwright/blob/e81a3c5901e024f7c41edf023daad60c68ed093d/src/utils/registry.ts#L238

This was probably a recently introduced bug which breaks the usage of Playwright on Vercel serverless functions. It worked a few months ago. On Playwright side a few users faced already into this bug.


mxschmitt:
Update: I went through the latest releases until 0.1.7 and it all of them the file was not added. Still not sure if its a bug in that repo or as-designed.

mxschmitt:
Yup will take a look, I keep you posted. 👍

mxschmitt:
I think I was able to simplify the test-case as small as it can get. Unfortunately I have to give away my notebook today and then get a new one in a few days since I'm switching jobs. Also I tried to dig into it and spend a few hours but was not sure about the internals how the matching works. Would be awesome if you could take over, with the small repro it should be easier to find the root cause. Thanks!

test/unit/asset-class-constructor-directory/input.js
```diff
@@ -0,0 +1,10 @@
+const fs = require("fs")
+const path = require("path")
+
+class Foobar {
+  constructor(packagePath) {
+    console.log(fs.readFileSync(path.join(packagePath, 'asset1.txt'), 'utf8'))
```
styfle, line 6: I tried changing `readFileSync` to `require` and that doesn't work either. So there must be something different about the npm package that made it work 🤔

Issue #185, opened 2021-04-14
Tracing fails with webpack 5 and concatenated path import


ijjk:
When building Next.js pages with webpack 5 the path import has been changed slightly which causes tracing to fail causing needed files to not be caught as expected. 

Example page output with webpack 5:

<details>

<summary>.next/server/pages/api/posts.js</summary>

```
(function() {
var exports = {};
exports.id = 223;
exports.ids = [223];
exports.modules = {

/***/ 216:
/***/ (function(__unused_webpack_module, __webpack_exports__, __webpack_require__) {

"use strict";

// EXPORTS
__webpack_require__.d(__webpack_exports__, {
  "ld": function() { return /* binding */ getSortedPostsData; }
});

// UNUSED EXPORTS: getAllPostIds, getPostData

;// CONCATENATED MODULE: external "fs"
var external_fs_namespaceObject = require("fs");;
var external_fs_default = /*#__PURE__*/__webpack_require__.n(external_fs_namespaceObject);
;// CONCATENATED MODULE: external "path"
var external_path_namespaceObject = require("path");;
var external_path_default = /*#__PURE__*/__webpack_require__.n(external_path_namespaceObject);
;// CONCATENATED MODULE: external "gray-matter"
var external_gray_matter_namespaceObject = require("gray-matter");;
var external_gray_matter_default = /*#__PURE__*/__webpack_require__.n(external_gray_matter_namespaceObject);
;// CONCATENATED MODULE: ./src/lib/posts.js
function ownKeys(object, enumerableOnly) { var keys = Object.keys(object); if (Object.getOwnPropertySymbols) { var symbols = Object.getOwnPropertySymbols(object); if (enumerableOnly) symbols = symbols.filter(function (sym) { return Object.getOwnPropertyDescriptor(object, sym).enumerable; }); keys.push.apply(keys, symbols); } return keys; }

function _objectSpread(target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i] != null ? arguments[i] : {}; if (i % 2) { ownKeys(Object(source), true).forEach(function (key) { _defineProperty(target, key, source[key]); }); } else if (Object.getOwnPropertyDescriptors) { Object.defineProperties(target, Object.getOwnPropertyDescriptors(source)); } else { ownKeys(Object(source)).forEach(function (key) { Object.defineProperty(target, key, Object.getOwnPropertyDescriptor(source, key)); }); } } return target; }

[...]

PR #186, opened 2021-04-14
fix: webpack 5 wrapper handling

guybedford:
Fixes https://github.com/vercel/nft/issues/185 adding support for the new Webpack 5 namespace form which is a function not an object with an `a` property. The fix supports both webpack 4 and webpack 5 in an indirect way by creating the wrapper `Object.assign(() => require('x'), { a: require('x') })`, with the expression entirely evaluated virtually within the static evaluation system as a way to inform the evaluation result.


styfle (approved):
Thanks!

ijjk (commented):
It looks like this still fails using the repo here https://github.com/sozonome/nextjs-webpack5-api-route-vercel-deployment-error and tracing `.next/server/pages/api/posts.js` after `yarn build`

The module cache/bootstrap code is also moved to a separate file in webpack 5 `webpack-runtime.js`, could this be causing a conflict here?

styfle (approved):

ijjk (approved):
Awesome, looks like it's catching it now! 🎉 🎉

PR #189, opened 2021-04-27
feat: allow await outside function

### LINKED ISSUE #188: Add support for `top-level await`
just noticed an Error in `warnings`:
```
Error: Failed to parse xxx.js as module:
Cannot use keyword 'await' outside an async function
    at Object.analyze [as default] ...
```

it already supported in node>=14.8.0, so i think there will be more and more projects using top-level await in the future.

hehex9:
closes #188

### DIFF
--- a/src/analyze.ts
+++ b/src/analyze.ts
@@ -239,7 +239,7 @@ export default async function analyze(id: string, code: string, job: Job): Promi
   let isESM = false;
 
   try {
-    ast = acorn.parse(code, { ecmaVersion: 2020, allowReturnOutsideFunction: true });
+    ast = acorn.parse(code, { ecmaVersion: 2020, allowReturnOutsideFunction: true, allowAwaitOutsideFunction: true });
     isESM = false;
   }
   catch (e) {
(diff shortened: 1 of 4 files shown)


guybedford (approved):

src/analyze.ts
```diff
@@ -239,7 +239,7 @@ export default async function analyze(id: string, code: string, job: Job): Promi
   let isESM = false;
 
   try {
-    ast = acorn.parse(code, { ecmaVersion: 2020, allowReturnOutsideFunction: true });
+    ast = acorn.parse(code, { ecmaVersion: 2020, allowReturnOutsideFunction: true, allowAwaitOutsideFunction: true });
```
guybedford, line 242: Actually this is incorrect - CommonJS modules do not support TLA.

```suggestion
    ast = acorn.parse(code, { ecmaVersion: 2020, allowReturnOutsideFunction: true, allowAwaitOutsideFunction: true });
```

guybedford (changes requested):
Let's just remove the CJS TLA case before merging.

guybedford:
We should add this functionality to ncc as well - https://github.com/vercel/webpack-asset-relocator-loader/blob/main/src/asset-relocator.js#L531.

guybedford (approved):

hehex9:
thanks 😉

> need a maintainer to approve running workflows

styfle (approved):
Thank you!

PR #195, opened 2021-05-08
fix: special case argon2 lib binding

### LINKED ISSUE #183: Add argon2 compatibility with vercel
Using argon2@0.27.2

Depoying this to vercel gives a an error in the serverless function tab saying cannot find or reference node module argon 2.  

https://vercel.com/support/articles/why-does-my-serverless-function-work-locally-but-not-when-deployed.
Based off this article, I believe the issue is due to argon2 native node-gyp requiring libraries.  And it seems vercel cannot support native to node  / node to native compiling.   

Only solution I see is if argon2 authors create an argon2js or javascript version with 0 dependencies like bcryptjs or vercel changes something about their network but me thinks vercel cannot do that by design with this serverless architecture.

Argon2 is the best open source cryptography available for those wanting to manually hash their own passwords.  And vercel network is the best for rapid development and hands free deployment.  It would be great if they could get along.

Any thoughts on how we solve this?

styfle:
Fixes #183

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "@types/node": "^14.14.37",
     "analytics-node": "^3.4.0-beta.1",
     "apollo-server-express": "^2.14.2",
-    "argon2": "^0.24.0",
+    "argon2": "^0.27.2",
     "auth0": "^2.27.1",
     "axios": "^0.21.1",
     "azure-storage": "^2.10.3",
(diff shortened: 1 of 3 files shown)


styfle:
@guybedford We might have an issue with `node-pre-gyp` deprecated in favor of `@mapbox/node-pre-gyp`

https://www.npmjs.com/package/node-pre-gyp

guybedford (approved):
Sounds like direct support for `@mapbox/node-pre-gyp` would be useful then?

PR #204, opened 2021-05-13
Repro: Cannot find module 'quirrel/vercel'

Skn0tt:
A Quirrel users is facing `Cannot find module 'quirrel/vercel'` when deploying to Vercel. I was able to track it down to being caused by `nft` not picking up the dependency, this PR adds a reproduction test for the behaviour.

Curious for your thoughts on this - is this an `nft` error, or should Quirrel change what files it publishes?

### DIFF
--- a/test/integration/quirrel.js
+++ b/test/integration/quirrel.js
@@ -1,0 +1,1 @@
+const { Queue } = require("quirrel/vercel")
--- a/package.json
+++ b/package.json
@@ -98,6 +98,7 @@
     "phantomjs-prebuilt": "^2.1.16",
     "polyfill-library": "3.93.0",
     "pug": "^3.0.1",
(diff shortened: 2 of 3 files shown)


guybedford:
@Skn0tt you need to include the `.cjs` or `.mjs` file extension for that subpath require to work in Node.js. Alternatively you can define exports with the "exports" field.

test/integration/quirrel.js
```diff
@@ -0,0 +1 @@
+const { Queue } = require("quirrel/vercel")
```
styfle, line 1: Perhaps you meant to use `require("quirrel/vercel.cjs") instead?

Skn0tt:
Thank you! Fixed the issue by publishing `.js` files in addition to `.cjs`.

Issue #201, opened 2021-05-11
Tracing fails to detect fs when used in multiple entries with webpack 5


ijjk:
Continuation of https://github.com/vercel/nft/issues/185

When building a Next.js project that reads from the filesystem in multiple pages the webpack output changes and `nft` fails to detect the needed local files. When only one page reads from the filesystem the webpack output is handled correctly. 

Related Next.js issue: https://github.com/vercel/next.js/issues/24700
Provided Next.js reproduction: https://github.com/jacksonhardaker/next-webpack-5-fs-bug

<details>

<summary>Example output when only one page reads from the filesystem (traces `/data` correctly)</summary>

```js
(function() {
var exports = {};
exports.id = 405;
exports.ids = [405];
exports.modules = {

/***/ 34:
/***/ (function(__unused_webpack_module, __webpack_exports__, __webpack_require__) {

"use strict";
// ESM COMPAT FLAG
__webpack_require__.r(__webpack_exports__);

// EXPORTS
__webpack_require__.d(__webpack_exports__, {
  "default": function() { return /* binding */ _; },
  "getServerSideProps": function() { return /* binding */ getServerSideProps; }
});

;// CONCATENATED MODULE: external "react/jsx-runtime"
var jsx_runtime_namespaceObject = require("react/jsx-runtime");;
;// CONCATENATED MODULE: external "fs"
var external_fs_namespaceObject = require("fs");;
;// CONCATENATED MODULE: external "path"
var external_path_namespaceObject = require("path");;
var external_path_default = /*#__PURE__*/__webpack_require__.n(external_path_namespaceObject);
;// CONCATENATED MODULE: ./pages/index.js




function _({
  text
}) {
  return /*#__PURE__*/(0,jsx_runtime_namespaceObject.jsxs)("div", {
    children: [/*#__PURE__*/jsx_runtime_namespaceObject.jsx("h1", {
      children: "next-webpack-5-fs-bug"
    }), /*#__PURE__*/jsx_runtime_namespaceObject.jsx("p", {
      children: text
    })]
  });
}
function getServerSideProps() {
  const text = (0,external_fs_namespaceObject.readFileSync)(external_path_default().resolve(process.cwd(), 'data/copy.txt'), {
    encoding: 'utf-8'
  });
  return {
    props: {
      text
    }
  };
}

/***/ })

};
;

// load runtime
var __webpack_require__ = require("../webpack-runtime.js");
__webpack_require__.C(exports);
[...]

PR #205, opened 2021-05-13
fix: support webpack chunked wrappers

guybedford:
Fixes Webpack wrapper detection to support non-parameter wrapper IIFEs and 'use strict' in external blocks.

Resolves https://github.com/vercel/nft/issues/201.


styfle (approved):
Thanks!

Issue #199, opened 2021-05-10
Add support for `@mapbox/node-pre-gyp`


styfle:
[node-pre-gyp](https://www.npmjs.com/package/node-pre-gyp) is deprecated in favor of [@mapbox/node-pre-gyp](https://www.npmjs.com/package/@mapbox/node-pre-gyp) so we need to add support for the new name.

PR #206, opened 2021-05-13
feat: support @mapbox/node-pre-gyp workflows

guybedford:
Resolves https://github.com/vercel/nft/issues/199.


src/analyze.ts
```diff
@@ -93,6 +95,10 @@ const staticModules = Object.assign(Object.create(null), {
     default: os,
     ...os
   },
+  '@mapbox/node-pre-gyp': {
+    default: mapboxPregyp,
+    ...mapboxPregyp
+  },
   'node-pre-gyp': pregyp,
   'node-pre-gyp/lib/pre-binding': pregyp,
   'node-pre-gyp/lib/pre-binding.js': pregyp,
```
styfle, line 104: Should we replace the legacy `node-pre-gyp` with the new mapbox one here or keep them both?

guybedford: We can keep both for legacy compat, possibly dropping the old one in a subsequent major.

styfle (approved):
Thanks again 👍

styfle:
I think we need to keep [`node-pre-gyp`](https://www.npmjs.com/package/node-pre-gyp) since it still gets 2.7 million downloads a week and has 812 packages that depend on it. 

Regarding `mkdirp` we can drop it in #260, thanks!

Issue #207, opened 2021-05-13
Missing CJS for dual-mode package `@rdf-esm/term-map`


styfle:
Steps to reproduce:

```sh
mkdir example
cd example
yarn init -y
yarn add @rdf-esm/term-map @tpluscode/sparql-builder
echo "require('@tpluscode/sparql-builder')" > index.js
node index.js        # works with standard node
nft build index.js
node dist/index.js   # fails after nft
```

Error:

```
Error: Cannot find module node_modules/@rdf-esm/term-map/index.cjs
```

PR #208, opened 2021-05-13
fix: exports resolution, fixup self resolution check

guybedford:
It turns out there was a bug in the self resolution check which was then stopping exports resolution from working out correctly in turn.

Resolves https://github.com/vercel/nft/issues/207.


src/resolve-dependency.ts
```diff
@@ -251,6 +251,7 @@ function resolvePackage (name: string, parent: string, job: Job, cjsResolve: boo
       }
     }
   }
+  console.log('[fallthrough]');
```
styfle, line 254: Let's remove this console.log()

styfle (approved):
Thanks!

PR #213, opened 2021-05-23
fix: add playwright-core support

### LINKED ISSUE #211: Cannot deploy Playwright to Vercel - required file not traced by NFT
Playwright cannot be deployed to Vercel since @vercel/nft 0.12.0 was released. This was previously working with @vercel/nft 0.11.2.

Playwright [requires a file named `browsers.json`](https://github.com/microsoft/playwright/blob/9cd89ae0525b327ce435ae3e4aadcba035621ee9/src/utils/registry.ts#L261), which is not being traced by @vercel/nft 0.12.0.

[This repo has a test case for both 0.11.2 (working) and 0.12.0 (not working)](https://github.com/juice49/vercel-nft-playwright-test).

This has also been discussed at microsoft/playwright#5862.

mxschmitt:
Verified that the test was failing without the special case and it's passing with the special case.

Closes #184
Closes #211

Relates https://github.com/microsoft/playwright/issues/5862

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "pdfkit": "^0.10.0",
     "pg": "^7.11.0",
     "phantomjs-prebuilt": "^2.1.16",
+    "playwright-core": "^1.11.1",
     "polyfill-library": "3.93.0",
     "pug": "^3.0.1",
     "react": "^16.8.6",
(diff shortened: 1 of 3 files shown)


styfle (approved):
Thanks!

PR #212, opened 2021-05-22
fix: wrappers cannot read type of undefined

styfle:
Fixes #210

I don't have the source code to reproduce so I didn't add a test

PR #214, opened 2021-05-25
fix: add missing @types/estree

styfle:
We've had several null/undefined bugs (such as #212) that can be avoided with better TS types.

This PR adds `@types/estree` as suggested by @sokra [a few years ago](https://github.com/acornjs/acorn/issues/741#issuecomment-426260204) and this seems to do the trick.

### DIFF
--- a/src/utils/wrappers.ts
+++ b/src/utils/wrappers.ts
@@ -1,6 +1,6 @@
 import { walk, Node as ESNode } from 'estree-walker';
 import { Ast } from '../types';
-import { SimpleCallExpression, Node, Literal, FunctionExpression, ReturnStatement, ObjectExpression, ArrayExpression, Property } from 'estree';
+import { SimpleCallExpression, Node, Literal, FunctionExpression, ReturnStatement, ObjectExpression, ArrayExpression, Property, CallExpression } from 'estree';
 
 function isUndefinedOrVoid (node: Node) {
   return node.type === 'Identifier' && node.name === 'undefined' || node.type === 'UnaryExpression' && node.operator === 'void' && node.argument.type === 'Literal' && node.argument.value === 0;
@@ -483,8 +483,9 @@ export function handleWrappers(ast: Ast) {
                   node.arguments[0].type === 'Literal') {
                 const externalId = externalMap.get(String(node.arguments[0].value));
                 if (externalId) {
-                  const replacement = {
+                  const replacement: CallExpression = {
                     type: 'CallExpression',
+                    optional: false,
                     callee: {
                       type: 'Identifier',
                       name: 'require'
@@ -495,19 +496,19 @@ export function handleWrappers(ast: Ast) {
                     }]
                   };
                   const parent = maybeParent!;
-                  if (parent.right === node) {
+                  if ('right' in parent && parent.right === node) {
                     parent.right = replacement;
                   }
-                  else if (parent.left === node) {
+                  else if ('left' in parent && parent.left === node) {
                     parent.left = replacement;
                   }
-                  else if (parent.object === node) {
+                  else if ('object' in parent && parent.object === node) {
                     parent.object = replacement;
                   }
-                  else if (parent.callee === node) {
+                  else if ('callee' in parent && parent.callee === node) {
                     parent.callee = replacement;
                   }
-                  else if (parent.arguments && parent.arguments.some(arg => arg === node)) {
+                  else if ('arguments' in parent && parent.arguments.some(arg => arg === node)) {
                     parent.arguments = parent.arguments.map(arg => arg === node ? replacement : arg);
                   }
                   else if ('init' in parent && parent.init === node) {
@@ -530,6 +531,7 @@ export function handleWrappers(ast: Ast) {
                   const req = node.arguments[0];
                   const callExpression: SimpleCallExpression = {
                     type: 'CallExpression',
+                    optional: false,
                     callee: {
                       type: 'MemberExpression',
                       computed: false,
@@ -553,7 +555,8 @@ export function handleWrappers(ast: Ast) {
                       {
                         type: 'ObjectExpression',
                         properties: [{
-                          type: 'ObjectProperty',
+                          type: 'Property',
+                          kind: 'init',
                           method: false,
                           computed: false,
                           shorthand: false,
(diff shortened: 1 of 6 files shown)


guybedford:
Nice work, let me know if you want help getting this over the line.

src/utils/wrappers.ts
```diff
@@ -164,24 +183,33 @@ export function handleWrappers(ast: Node) {
...
-          const cache = (wrapper.arguments[0].body.body[1] || wrapper.arguments[0].body.body[0]).argument.callee.arguments[1];
+        const argBody = (wrapper.arguments[0].body.body[1] || wrapper.arguments[0].body.body[0]);
+        if (externalIds.length &&
+            argBody &&
+            'argument' in argBody &&
+            argBody.argument &&
+            'callee' in argBody.argument &&
+            'arguments' in argBody.argument.callee) {
+          const cache = argBody.argument.callee.arguments[1];
           cache.properties = externalIds.map(ext => {
```
styfle, line 213: @guybedford Is there a certain `cache.type` you expect here? Not all of the types have `properties` which is why its complaining.

src/utils/wrappers.ts
```diff
@@ -386,21 +424,27 @@ export function handleWrappers(ast: Node) {
...
+        throw new Error('Expected ExpressionStatement');
+      }
+      const moduleObj = wrapper.callee.params.length ? wrapper.arguments[0] : statement.expression.right;
+      let modules: [number, typeof moduleObj][];
       if (moduleObj.type === 'ArrayExpression')
-        modules = moduleObj.elements.map((el: any, i: number) => [i, el]);
+        modules = moduleObj.elements.map((el, i) => [i, el]);
       else
-        modules = moduleObj.properties.map((prop: any) => [prop.key.value, prop.value]);
+        modules = 'properties' in moduleObj ? moduleObj.properties.map(prop => [prop.key.value, prop.value]) : [];
```
styfle, line 447: @guybedford I added the check for existence of `properties` but the remaining problem here is that `prop.key` might not exist. Should I Filter out all the items where `prop.key` doesnt exist?

src/utils/wrappers.ts
```diff
@@ -386,21 +424,27 @@ export function handleWrappers(ast: Node) {
...
-      let modules: [number, any][];
+      const statement = wrapper.callee.body.body[3];
+      if (statement.type !== 'ExpressionStatement' || statement.expression.type !== 'AssignmentExpression' || statement.expression.right.type !== 'ObjectExpression') {
+        throw new Error('Expected ExpressionStatement');
+      }
+      const moduleObj = wrapper.callee.params.length ? wrapper.arguments[0] : statement.expression.right;
+      let modules: [number, typeof moduleObj][];
       if (moduleObj.type === 'ArrayExpression')
-        modules = moduleObj.elements.map((el: any, i: number) => [i, el]);
+        modules = moduleObj.elements.map((el, i) => [i, el]);
```
styfle, line 445: This is complaining because `el` can be null. Should I filter those out?

src/utils/wrappers.ts
```diff
@@ -386,21 +424,27 @@ export function handleWrappers(ast: Node) {
...
           wrapper.callee.body.body[5].declarations[0].init.callee.type === 'Identifier' &&
           wrapper.callee.body.body[5].declarations[0].init.callee.name === 'require')) {
       const externalMap = new Map<number, any>();
-      const moduleObj = wrapper.callee.params.length ? wrapper.arguments[0] : wrapper.callee.body.body[3].expression.right;
-      let modules: [number, any][];
+      const statement = wrapper.callee.body.body[3];
+      if (statement.type !== 'ExpressionStatement' || statement.expression.type !== 'AssignmentExpression' || statement.expression.right.type !== 'ObjectExpression') {
+        throw new Error('Expected ExpressionStatement');
+      }
+      const moduleObj = wrapper.callee.params.length ? wrapper.arguments[0] : statement.expression.right;
```
styfle, line 442: I'm wondering if the bug is here because `wrapper.arguments[0]` can basically be anything but `statement.expression.right` can only be `ObjectExpression`.

Do we need to narrow down `wrapper.arguments[0]` to a specific type?

src/utils/wrappers.ts
```diff
@@ -164,24 +183,33 @@ export function handleWrappers(ast: Node) {
...
+        if (externalIds.length &&
+            argBody &&
+            'argument' in argBody &&
+            argBody.argument &&
+            'callee' in argBody.argument &&
+            'arguments' in argBody.argument.callee) {
+          const cache = argBody.argument.callee.arguments[1];
           cache.properties = externalIds.map(ext => {
             return {
               type: 'Property',
```
styfle, line 215: This whole return type is complaining because it doesn't match Property type. Its missing the following:

```
method: boolean;
shorthand: boolean;
computed: boolean;
```

Also, the nested `CallExpression` is missing the `optional: boolean`.

Is it safe to say these should all be `false`?

guybedford:
I've pushed up the type fixes here - CI should be passing now.

src/utils/wrappers.ts
```diff
@@ -427,16 +469,19 @@ export function handleWrappers(ast: Node) {
...
       for (const [, m] of modules) {
-        if (m.params.length === 3 && m.params[2].type === 'Identifier') {
+        if ('params' in m && m.params.length === 3 && m.params[2].type === 'Identifier') {
           const assignedVars = new Map();
-          walk(m.body, {
-            enter (node, maybeParent) {
+          walk((m.body as unknown) as ESNode, {
+            enter (_node, _maybeParent) {
+              const node = _node as Node;
+              const maybeParent = _maybeParent as Node | undefined;
```
styfle, line 477: @guybedford What are your thoughts here? I'm pretty sure this is supposed to be a `Node` not `ESNode` and its surfacing a couple bugs below.

guybedford: The `walk` function here uses a different `Node` base type that has the `start` and `end` requirements as well. Typing this function properly is thus difficult due to the compat differences - can't use the base typing library you need to use another one.

styfle: I’m not so much interested in the walk function but the enter handler. Can you talk a look at my most recent commit here to see how to resolve the error below?

guybedford: Sure, I've pushed up a fix.

PR #215, opened 2021-06-01
feat: add cli command `size`

styfle:
Instead of `nft build index.js && du -sh dist`, this introduces a new command: `nft size index.js`.

I also updated the CLI output to be a little friendlier:

```
△ nft 0.12.2

Usage:

  $ nft [command] <file>

Commands:

  build    trace and copy to the dist directory
  print    trace and print to stdout
   size    trace and print size in bytes
```

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -16,7 +16,7 @@ async function cli(
   const opts = {
     ts: true,
     mixedModules: true,
-    log: true
+    log: action !== 'size'
   };
 
   const { fileList, esmFileList, warnings } = await nodeFileTrace(files, opts);
@@ -52,7 +52,7 @@ async function cli(
         bytes += stat.size;
       }
     }
-    stdout.push(`${(bytes / 1e6).toFixed(2)} MB`)
+    stdout.push(`${bytes} bytes total`)
   } else {
     stdout.push(`△ nft ${require('../package.json').version}`);
     stdout.push('');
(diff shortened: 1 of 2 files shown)


src/cli.ts
```diff
@@ -39,8 +40,31 @@ async function cli(
...
+    for (const f of allFiles) {
+      const lstat = lstatSync(f);
+      if (isSymbolicLink(lstat.mode)) {
+        bytes += lstat.size;
+      } else {
+        const stat = statSync(f)
+        bytes += stat.size;
+      }
+    }
+    stdout.push(`${bytes} bytes total`)
```
guybedford, line 55: It could be preferable to make this KB by default, or choose between bytes or not. Here's an example function - https://stackoverflow.com/questions/10420352/converting-file-size-in-bytes-to-human-readable-string/14919494#14919494.

guybedford (approved):
Great addition

src/cli.ts
ijjk: We could potentially leverage [`pretty-bytes`](https://www.npmjs.com/package/pretty-bytes) here

styfle: I started going down that road but I wasn't sure if we wanted to use base 2 (1024) or base 10 (1000) to mean 1KB (I'll need to verify how the API does it) so I left that for a follow up PR

guybedford: Perhaps at least use numeric separators with a standard format like `123,456,789B`?

ijjk (approved):

PR #224, opened 2021-06-10
feat: private fields support

guybedford:
Adds the acorn plugin for private fields and removes the ignored ecmascript tests.

It turns out a lot of the previously ignored tests are no longer present as well so there are currently no ignores!


styfle (approved):
Wow, that's great!

PR #228, opened 2021-08-07
fix: add playwright support

ThewBear:
Similar to #213 but for `playwright` package.


ThewBear:
No longer in used since the [50mb function deploy limit](https://vercel.link/serverless-function-size) make it impossible to include the browser binary.

PR #232, opened 2021-09-15
Add special case for `remark-prism`

styfle:
This PR adds a special case since `nft` can't statically analyze this dynamic `require.resolve()`:

https://github.com/sergioramos/remark-prism/blob/49f1f86d50f376c647214c02cf972ce05b0fc756/src/highlight.js#L74

```js
  getLoader(components, languages, loaded).load((lang) => {
    if (!(lang in components.languages)) {
      console.warn('Language does not exist: ' + lang);
      return;
    }

    const filename = require.resolve(`prismjs/components/prism-${lang}`);
    runInContext(RUN(readFileSync(filename, 'utf-8')), parsingContext, {
      filename,
    });

    loadedLanguages.add(lang);
  });
```


src/utils/special-cases.ts
```diff
@@ -82,6 +82,17 @@ const specialCases: Record<string, (o: SpecialCaseOpts) => void> = {
       emitAssetDirectory(resolve(dirname(id), '..', 'bin'));
     }
   },
+  'remark-prism' ({ id, emitAsset }) {
+    const file = 'remark-prism/src/highlight.js';
+    if (id.endsWith(file)) {
+      try {
+        const prefix = id.slice(0, -file.length);
+        emitAsset(resolve(prefix, `prismjs/components/prism-markdown.js`));
```
ijjk, line 90: Should this only be adding `prism-markdown` or should we be including all of the files in the components folder here since we won't know which are needed?

styfle: Yeah good point! Added in 43d6a45 

I also added an integration test in 85b7608

ijjk (approved):

PR #236, opened 2021-09-22
Convert fs operations from sync to async

ijjk:
This converts our sync fs usage to async to help with performance and to allow external `resolve` functions to be async as well. Noticed the synchronous resolving was taking a big chunk of time while tracing during `next build` so hopefully this allows us to hook into webpack's resolver and give a good performance boost.


src/analyze.ts
```diff
@@ -19,6 +18,8 @@ import mapboxPregyp from '@mapbox/node-pre-gyp';
 import { Job } from './node-file-trace';
 import { fileURLToPath, pathToFileURL, URL } from 'url';
 
+const asyncWalk: typeof import('../node_modules/estree-walker').asyncWalk = require('estree-walker').asyncWalk
```
styfle, line 21: Why is this using require instead of import?

src/node-file-trace.ts
```diff
@@ -19,19 +25,19 @@ export async function nodeFileTrace(files: string[], opts: NodeFileTraceOptions
   const job = new Job(opts);
 
   if (opts.readFile)
-    job.readFile = opts.readFile;
+    job.readFile = opts.readFile as typeof job.readFile;
```
styfle, line 28: Can we do this without “as”. Something seems wrong with the types

src/types.ts
```diff
@@ -45,10 +45,10 @@ export interface NodeFileTraceOptions {
...
   log?: boolean;
   mixedModules?: boolean;
-  readFile?: (path: string) => Buffer | string | null;
-  stat?: (path: string) => Stats | null;
-  readlink?: (path: string) => string | null;
-  resolve?: (id: string, parent: string, job: Job, cjsResolve: boolean) => string | string[];
+  readFile?: (path: string) => Buffer | string | null | Promise<Buffer | string | null>;
+  stat?: (path: string) => Stats | null | Promise<Stats | null>;
+  readlink?: (path: string) => string | null | Promise<Stats | null>;
+  resolve?: (id: string, parent: string, job: Job, cjsResolve: boolean) => string | string[] | Promise<string | string[]>;
```
styfle, line 51: If you’re worried about a breaking change, I would be fine with making these all promise return types since it’s only braking TS users. Awaiting a sync function will still work for example.

src/utils/is-error.ts
```diff
@@ -0,0 +1,11 @@
+// We allow some additional attached properties for Errors
+interface NftError {
+  code?: string | number
+  message?: string
+}
+
+export default function isError(err: unknown): err is NftError {
+  return (
+    typeof err === 'object' && err !== null && ('code' in err || 'message' in err)
```
styfle, line 9: Isn’t type of Error “function”?  

CC @mrmckeb for helpers

src/analyze.ts
ijjk: TypeScript was resolving to the `estree-walker` version used in `rollup-pluginutils` instead of the top-level `estree-walker` version which doesn't have `asyncWalk` so it's to fix the incorrect type

src/types.ts
ijjk: Ah ok yeah was worried about unexpected type change, will update to just returning Promises

src/utils/is-error.ts
ijjk: I copied/modified this helper from the Next.js suite here https://github.com/vercel/next.js/blob/f8e283c7ba9f8f684511c00aee5e6742123dd2ba/packages/next/lib/is-error.ts

src/types.ts
ijjk: Updated in https://github.com/vercel/nft/pull/236/commits/66bee8668c17d65dacf39bc05002c9943b97386f

src/utils/is-error.ts
styfle: Oh my bad, I tried `console.log(typeof Error)` but that should have been `console.log(typeof new Error())` which is the expected function.

That said, we should probably use the helpers that @mrmckeb wrote for the api.

src/analyze.ts
styfle: Lets add a comment why this is here because we're going to forget 👍

styfle (commented):
I think you should add a test for async usage too 🙂

src/utils/is-error.ts
styfle: Let's revisit this when we update `typescript` in this repo, that can be another PR

ijjk: Updated to remove these type checks since they aren't actually needed yet as discussed

src/analyze.ts
ijjk: Added in https://github.com/vercel/nft/pull/236/commits/02db8aeb263ff2318dbe162c64c0fd45480df7e2

test/unit.test.js
```diff
@@ -17,8 +17,51 @@ for (const { testName, isRoot } of unitTests) {
     console.log(`Skipping unit test on Windows: ${testSuffix}`);
     continue;
   };
+  const unitPath = join(__dirname, 'unit', testName);
+  
+  if (testName === 'wildcard') {
+    it('should work with async fs options used', async () => {
```
styfle, line 23: How about updating line 98 instead of a whole new test? Something like:

```diff
- readFile: readFileMock,
+ readFile: testName === 'someAsyncTest' ? readFileMockAsync : readFileMock,
```

ijjk: Updated in https://github.com/vercel/nft/pull/236/commits/e04da34f55a34693ae01dba2dfda54a18f0cf56c

styfle (approved):
Nice work, thanks! 🎉

PR #237, opened 2021-10-01
Add additional items to shared cache

ijjk:
This ensures we leverage the shared cache when preforming expensive checks to provide better performance when tracing multiple files individually but sharing the same cache instead of tracing them in the same `nodeFileTrace` run. 

In one test repo this brought tracing down from 130 seconds -> 25 seconds while tracing separately. which brings this much closer with the time it took when tracing all files in one `nodeFileTrace` run which was around 17 seconds.

### DIFF
--- a/src/node-file-trace.ts
+++ b/src/node-file-trace.ts
@@ -3,7 +3,7 @@ import { basename, dirname, extname, relative, resolve, sep } from 'path';
 import fs from 'fs';
 import { promisify } from 'util'
 import analyze, { AnalyzeResult } from './analyze';
-import resolveDependency from './resolve-dependency';
+import resolveDependency, { FilesToEmit } from './resolve-dependency';
 import { isMatch } from 'micromatch';
 import { sharedLibEmit } from './utils/sharedlib-emit';
 import { join } from 'path';
@@ -237,7 +237,7 @@ export class Job {
     }
   }
 
-  async realpath (path: string, parent?: string, seen = new Set()): Promise<string> {
+  async realpath (path: string, parent?: string, seen = new Set(), filesToEmit ?: FilesToEmit): Promise<string> {
     if (seen.has(path)) throw new Error('Recursive symlink detected resolving ' + path);
     seen.add(path);
     const symlink = await this.readlink(path);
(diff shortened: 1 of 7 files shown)


src/node-file-trace.ts
```diff
@@ -270,13 +284,20 @@ export class Job {
   }
 
   async getPjsonBoundary (path: string) {
+    const initialPath = path
+    const cacheItem = this.pjsonBoundaryCache.get(path)
+    if (typeof cacheItem !== 'undefined') return cacheItem
```
guybedford, line 289: This cache may be redundant since the only FS operations this method depends on is `isFile` which uses the stat cache, which uses `null` for empty entries so does fully cache both existing and non existing package.json files already. The loop saving should be negligible since file depths usually aren't that large.

So for simplicity it may well be worth removing this `pjsonBoundaryCache` and just relying on the stat cache - there shouldn't be any difference in performance.

src/resolve-dependency.ts
```diff
@@ -1,42 +1,72 @@
...
+  type: string,
+  parent?: string,
+  isRealpath?: boolean
+}>
+
 // node resolver
 // custom implementation to emit only needed package.json files for resolver
 // (package.json files are emitted as they are hit)
 export default async function resolveDependency (specifier: string, parent: string, job: Job, cjsResolve = true): Promise<string | string[]> {
+  const cacheKey = specifier + parent
```
guybedford, line 15: Nice improvement, I'm sure this saves a lot of work.

The cache key should take into account the environment conditions as well here. Also it's advisable to properly terminate. Eg something like:

```js
const { conditions } = job;
const cacheKey = JSON.stringify({ specifier, parent, conditions })
```

guybedford (approved):

src/resolve-dependency.ts
ijjk: Good catch, updated to include that

src/node-file-trace.ts
ijjk: Yeah makes sense, double checked and it looks like it doesn't provide any difference in performance so removed it to reduce the additional complexity.

PR #238, opened 2021-10-09
Add emitDependency caching and fix pnpm symlinks

ijjk:
This removes the `resolve` caching introduced in https://github.com/vercel/nft/pull/237 and replaces it with caching the entire `emitDependency` call so that all nested operations shouldn't need to be cache separately. 

This provides much better performance than the previous caching as testing against a couple large apps showed these results:

- vercel.com: before(70 - 100 seconds) after(20 - 30 seconds)
- private project: before(170 - 200 seconds) after(50 - 60 seconds)


src/node-file-trace.ts
```diff
@@ -277,106 +276,155 @@ export class Job {
...
-        }
-        if (Array.isArray(resolved)) {
-          for (const item of resolved) {
-            // ignore builtins
-            if (item.startsWith('node:')) return;
-            await this.emitDependency(item, path);
+    const emitDependencyPromise = new Promise<FilesToEmit>(async (resolve, reject) => {
+      try {
+        const curFilesToEmit: FilesToEmit = { files: [], reasons: {} }
+        const job = new Proxy(this, {
```
styfle, line 309: Why use a proxy here?

guybedford:
Great improvements, and looks good to me, didn't read all the code but could just ask a few questions if it would help:

* How the gains are achieved here? If it is primarily disk, what were the misses on the previous PR? And if it is primarily due to compute, that would be very surprising since 1000s of dependencies should be fully iterable in well under 5 seconds.
* With the caching stuff it's also worth putting some thought to lifetimes - what's nice about the stat cache etc is they are easier to mutate / coordinate on for longer period caching, whereas approaches like these PRs move towards more opaque caching. Or possibly doing a separation of node_modules caching from user code caching. What sort of cache lifetime applies to these types of builds?

guybedford:
> It seems eliminating iterating over the dependencies reduces the time significantly on large projects e.g. thousands of dependencies with > 100 separate runs of nodeFileTrace sharing the same cache.

This really is very surprising to me, as it is fully possible to iterate 1000s of modules in well under a second. So there must be something else going on. Could be worth digging into some further analysis there, but this PR is good to land too of course.

test/unit/pnpm-symlinks/output.js
```diff
@@ -0,0 +1,11 @@
+[
+  "test/unit/pnpm-symlinks/input.js",
+  "test/unit/pnpm-symlinks/node_modules/.pnpm/parse5-htmlparser2-tree-adapter@6.0.1/node_modules/parse5",
```
pi0, line 3: Is this an expected behavior that includes symlink directories as NodeFileTraceResult?

PR #239, opened 2021-10-13
Fix caching with circular emitDependency

ijjk:
Follow-up to https://github.com/vercel/nft/pull/238 this ensures we handle caching for circular `emitDependency` calls correctly. This also enables testing of the shared caching against the integration suite as it was previously only being tested against the unit suite.

### DIFF
--- a/src/node-file-trace.ts
+++ b/src/node-file-trace.ts
@@ -294,13 +294,34 @@ export class Job {
   }
 
   async emitDependency (path: string, parent?: string, filesToEmit?: FilesToEmit) {
-    if (this.processed.has(path)) return;
-    this.processed.add(path);
-
+    if (this.processed.has(path) && this.emitDependencyCache.has(path)) {
+      const cacheItem = this.emitDependencyCache.get(path)
+      
+      if (filesToEmit && cacheItem) {
+        this.emitDependencyCache.set(path, cacheItem.then(res => {
+          res.files.forEach(file => filesToEmit.files.push(file))
+          Object.assign(filesToEmit.reasons, res.reasons)
+          return res
+        }))
+      }
+      return
+    }
+    this.processed.add(path)
+    
     if (this.emitDependencyCache.has(path)) {
       const toEmit = await this.emitDependencyCache.get(path)!
+      
       toEmit.files.forEach(file => this.fileList.add(file))
       Object.assign(this.reasons, toEmit.reasons)
+      
+      if (filesToEmit) {
+        toEmit.files.forEach(file => {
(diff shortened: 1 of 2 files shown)


src/node-file-trace.ts
```diff
@@ -294,13 +294,34 @@ export class Job {
   }
 
   async emitDependency (path: string, parent?: string, filesToEmit?: FilesToEmit) {
-    if (this.processed.has(path)) return;
```
styfle, line 297: This early return changed.

Do we need to consider the case when `this.processed.has(path)` is true but `this.emitDependencyCache.has(path)` is false?

src/node-file-trace.ts
```diff
@@ -294,13 +294,34 @@ export class Job {
...
-    this.processed.add(path);
-
+    if (this.processed.has(path) && this.emitDependencyCache.has(path)) {
+      const cacheItem = this.emitDependencyCache.get(path)
```
styfle, line 298: Let's move 298 above 297 so that we can get and check for existence in the same line.

(not sure if there is a scenario where we remove from emitDependencyCache)

src/node-file-trace.ts
ijjk: That shouldn't happen as they are both set the first time `emitDependency` is called for a path, this check was more so for TypeScript initially, we can remove it if it makes it less clear I think since we're checking `cacheItem` in the inner if already.

ijjk: Updated in https://github.com/vercel/nft/pull/239/commits/00d1f5871680ca3b0a8a67d3484cecd04a474dbf

src/node-file-trace.ts
```diff
@@ -294,13 +294,34 @@ export class Job {
...
-    if (this.processed.has(path)) return;
-    this.processed.add(path);
-
-    if (this.emitDependencyCache.has(path)) {
-      const toEmit = await this.emitDependencyCache.get(path)!
+    const cacheItem = this.emitDependencyCache.get(path)
+    
+    if (this.processed.has(path)) {
+      if (filesToEmit && cacheItem) {
+        this.emitDependencyCache.set(path, cacheItem.then(res => {
```
styfle, line 301: Any reason why you're using `cacheItem.then()` instead of await cacheItem`?

src/node-file-trace.ts
```diff
@@ -294,13 +294,34 @@ export class Job {
...
+          res.files.forEach(file => filesToEmit.files.push(file))
```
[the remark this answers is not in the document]
ijjk: Yeah, good catch

PR #240, opened 2021-10-15
Remove emit cache in favor of ensuring all parents are added in reasons

ijjk:
This removes the emit caching as it seems there are cases where files aren't being emitted properly and adds complexity to tracing. The main problem trying to be addressed is to be able to collect specific traced files for a specific input without incurring a performance hit. 

Since tracing the input files together provides better performance than tracing separately as work doesn't need to be redone nearly as much even with the cache this ensures the reasons is updated properly with all parents so that we can walk that and collect the specific files for a specific input instead. 

As discussed with @styfle this also removes returning sorted arrays from `nodeFileTrace` in favor of returning the `Set`s directly and converts the `reasons` to a `Map`.

### DIFF
--- a/test/integration.test.js
+++ b/test/integration.test.js
@@ -55,14 +55,18 @@ for (const integrationTest of readdirSync(integrationDir)) {
     expect(code).toBe(fails ? 1 : 0);
     rimraf.sync(tmpdir);
     
-    const cachedResult = await nodeFileTrace([`${integrationDir}/${integrationTest}`], {
-      log: true,
-      cache: nftCache,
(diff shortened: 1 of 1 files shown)


src/node-file-trace.ts
```diff
@@ -236,50 +228,55 @@ export class Job {
...
     }
+    if (parent) {
+      reasonEntry.parents.add(parent);
+    }
     this.fileList.add(path);
-    return {
-      path,
-      reasonEntry
-    };
+    return true;
```
styfle, line 279: Why is this returning true instead of the emitted file?

PR #243, opened 2021-10-26
feat: Add support for Node.js 16

styfle:
This PR adds support for [Node.js 16 LTS](https://github.com/nodejs/node/releases/tag/v16.13.0) by bumping a few dependencies. 

We also dropped Node.js 12 from CI but support is still there since tsconfig.json is unmodified.

In the future, we can update tsconfig.json target to es2020 once Node.js 12 reaches EOL.


ijjk (approved):

src/analyze.ts
```diff
@@ -25,9 +25,9 @@ const asyncWalk: typeof import('../node_modules/estree-walker').asyncWalk = requ
 
 // Note: these should be deprecated over time as they ship in Acorn core
 const acorn = Parser.extend(
-  require("acorn-class-fields"),
-  require("acorn-static-class-features"),
-  require("acorn-private-class-elements")
+  //require("acorn-class-fields"),
```
ijjk, line 28: Should we remove these dependencies now? https://github.com/vercel/nft/blob/e89f3ca8cfa41838d0c4d7c16255eba7f561609f/package.json#L24-L26

src/analyze.ts
```diff
@@ -25,9 +25,9 @@ const asyncWalk: typeof import('../node_modules/estree-walker').asyncWalk = requ
...
+  //require("acorn-static-class-features"),
+  //require("acorn-private-class-elements")
```
ijjk, line 30: Should we fully remove these lines as tests are passing with them disabled?

styfle: I'm going to do a follow up PR to make sure we support all Node.js 16 features and therefore plugins

PR #262, opened 2022-01-21
Add special case for `esbuild`

### LINKED ISSUE #246: Usage of "esbuild" package fails to find it's os-specific counterpart.
I would expect that if I was using a linux machine that and requiring "esbuild" into my project that "esbuild-linux-64" would also be listed as a needed dependency.

styfle:
- Fixes #246 
- Related to https://github.com/vercel/vercel/issues/7287
- Related to https://github.com/vercel/next.js/issues/33294

### DIFF
--- a/src/utils/special-cases.ts
+++ b/src/utils/special-cases.ts
@@ -29,6 +29,16 @@ const specialCases: Record<string, (o: SpecialCaseOpts) => void> = {
       emitAsset(resolve(dirname(id), 'camaro.wasm'));
     }
   },
+  'esbuild' ({ id, emitAssetDirectory }) {
+    if (id.endsWith('esbuild/lib/main.js')) {
+      const file = resolve(id, '..', '..', 'package.json');
+      const pkg = JSON.parse(readFileSync(file, 'utf8'));
+      for (const dep of Object.keys(pkg.optionalDependencies || {})) {
+        const dir = resolve(id, '..', '..', '..', dep);
+        emitAssetDirectory(dir);
+      }
+    }
+  },
   'google-gax' ({ id, ast, emitAssetDirectory }) {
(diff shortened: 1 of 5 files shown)


src/utils/special-cases.ts
```diff
@@ -29,6 +29,16 @@ const specialCases: Record<string, (o: SpecialCaseOpts) => void> = {
       emitAsset(resolve(dirname(id), 'camaro.wasm'));
     }
   },
+  'esbuild' ({ id, emitAssetDirectory }) {
+    if (id.endsWith('esbuild/lib/main.js')) {
+      const file = resolve(id, '..', '..', 'package.json');
+      const pkg = JSON.parse(readFileSync(file, 'utf8'));
+      for (const dep of Object.keys(pkg.optionalDependencies || {})) {
+        const dir = resolve(id, '..', '..', '..', dep);
+        emitAssetDirectory(dir);
```
ijjk, line 38: Is this emitting all the dependencies for all platforms/architectures? Are there any signals we can use to tell which should be included and which not or would the user need to use `ignore` to filter it down manually?

styfle: Only one of the optional dependencies is installed so it will never emit all of them as far as I can tell because they wont all exist

ijjk: Ah ok so this requires tracing on the same system you are deploying to so that the platform/architecture match up?

styfle: Right but the problem you're describing is not so much a tracing problem as it is a build problem. You always need to build on the same "platform" that you deploy to – thats why vercel.com is so nice because it builds and deploys 🙂

ijjk (approved):

PR #280, opened 2022-03-19
chore: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) or [String.prototype.substring()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) (if the second parameter could be negative) which works similarily but aren't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/src/analyze.ts
+++ b/src/analyze.ts
@@ -219,8 +219,8 @@ export default async function analyze(id: string, code: string, job: Job): Promi
     if (!job.analysis.emitGlobs) return;
     const wildcardIndex = wildcardPath.indexOf(WILDCARD);
     const dirIndex = wildcardIndex === -1 ? wildcardPath.length : wildcardPath.lastIndexOf(path.sep, wildcardIndex);
-    const assetDirPath = wildcardPath.substr(0, dirIndex);
-    const patternPath = wildcardPath.substr(dirIndex);
+    const assetDirPath = wildcardPath.substring(0, dirIndex);
+    const patternPath = wildcardPath.slice(dirIndex);
     const wildcardPattern = patternPath.replace(wildcardRegEx, (_match, index) => {
       return patternPath[index - 1] === path.sep ? '**/*' : '*';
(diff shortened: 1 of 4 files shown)


src/analyze.ts
```diff
@@ -389,8 +389,8 @@ export default async function analyze(id: string, code: string, job: Job): Promi
 
     const wildcardIndex = wildcardRequire.indexOf(WILDCARD);
     const dirIndex = wildcardIndex === -1 ? wildcardRequire.length : wildcardRequire.lastIndexOf(path.sep, wildcardIndex);
-    const wildcardDirPath = wildcardRequire.substr(0, dirIndex);
-    const patternPath = wildcardRequire.substr(dirIndex);
+    const wildcardDirPath = wildcardRequire.substring(0, dirIndex);
+    const patternPath = wildcardRequire.slice(dirIndex);
```
styfle, line 393: Why are some substring and others slice?

CommanderRoot: My intend was to copy the behaviour of the current `substr()`

`slice()` and `substring()` behave differently if the second parameter is negative. 
Examples
```
"test".slice(0, -1) == "tes"
"test".substring(0, -1) == ""
"test".substr(0, -1) == ""
```

Sometimes you want to remove the last x chars in which case `slice()` is what you want but if the second parameter can be positive or negative it's better to use substring* (if the first param is 0, if not it can [depend](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#differences_between_substring_and_slice) as well)

styfle (approved):

PR #281, opened 2022-04-11
fix: add pixelmatch to special case list

Ethan-Arrowood:
Adds unit and integration tests for the `pixelmatch` package which requires a special case for including the `bin` directory to correctly trace the dependency `pngjs`

Fixes a playwright bug https://github.com/microsoft/playwright/blob/7f513360681d2517caa24d266def9b3a7be4f469/packages/playwright-core/src/utils/comparators.ts#L23-L24


styfle (approved):
Thanks!

styfle:
For posterity: this fixes a playwright bug:

https://github.com/microsoft/playwright/blob/7f513360681d2517caa24d266def9b3a7be4f469/packages/playwright-core/src/utils/comparators.ts#L23-L24

PR #285, opened 2022-05-17
fix: handle `node:` prefix

### LINKED ISSUE #284: `node:` prefixes cause a warning
Running `nodeFileTrace` on an app that imports Node built-ins with the `node:` prefix (which is the preferred way to import built-ins as of Node 18) causes a warning to be emitted:

```js
import fs from 'node:fs';
```

```
warnings: Set(1) {
  Error: Failed to resolve dependency node:fs:
  Cannot find module 'node:fs' loaded from [dir]/src/index.js
      at [dir]/node_modules/@vercel/nft/out/node-file-trace.js:304:39
      at async Promise.all (index 0)
      at async Job.emitDependency ([dir]/node_modules/@vercel/nft/out/node-file-trace.js:290:9)
      at async Promise.all (index 0)
      at async nodeFileTrace ([dir]/node_modules/@vercel/nft/out/node-file-trace.js:32:5)
      at async file://[dir]/index.js:3:13
```

Removing the `node:` prefix fixes it, but that's not always an option since it could be in library code.

Reproduction here: https://github.com/Rich-Harris/vercel-nft-repro

Rich-Harris:
This appears to be enough to fix #284, but I'm not sure where a test would go

### DIFF
--- a/src/analyze.ts
+++ b/src/analyze.ts
@@ -302,7 +302,7 @@ export default async function analyze(id: string, code: string, job: Job): Promi
         value: {
           [FUNCTION] (specifier: string) {
             deps.add(specifier);
-            const m = staticModules[specifier];
+            const m = staticModules[specifier.startsWith('node:') ? specifier.slice(5) : specifier];
             return m.default;
           },
           resolve (specifier: string) {
@@ -342,7 +342,7 @@ export default async function analyze(id: string, code: string, job: Job): Promi
       if (decl.type === 'ImportDeclaration') {
         const source = String(decl.source.value);
         deps.add(source);
-        const staticModule = staticModules[source];
+        const staticModule = staticModules[source.startsWith('node:') ? source.slice(5) : source];
         if (staticModule) {
           for (const impt of decl.specifiers) {
             if (impt.type === 'ImportNamespaceSpecifier')
--- a/src/resolve-dependency.ts
+++ b/src/resolve-dependency.ts
@@ -139,7 +139,7 @@ function resolveExportsImports (pkgPath: string, obj: PackageTarget, subpath: st
   } else {
     matchObj = obj;
   }
-  
+
   if (subpath in matchObj) {
     const target = getExportsTarget(matchObj[subpath], job.conditions, cjsResolve);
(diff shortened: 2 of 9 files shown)


Rich-Harris:
Looks like that didn't fix it after all 😬

guybedford:
@Rich-Harris let me know if I can assist with any NFT stuff anytime - it seems like the unit test just needs to be updated? The logging isn't ideal, may make sense to log the error before the warning to ensure the error can be seen at that call site in https://github.com/vercel/nft/blob/main/test/unit.test.js#L175.

Rich-Harris:
@guybedford if you are able to get this working, that would be terrific — unfortunately I'm maxed out at the moment otherwise I'd spend more time digging into it. Right now I'm just [working around it](https://github.com/sveltejs/kit/blob/88bc0b0e814490e526f2297ea13f01d278158540/packages/adapter-vercel/index.js#L412-L416):

```js
	traced.warnings.forEach((error) => {
		// pending https://github.com/vercel/nft/issues/284
		if (error.message.startsWith('Failed to resolve dependency node:')) return;
		console.error(error);
	});
```

guybedford:
I don't have time to clone and PR right now, especially since I'm on Windows here. But I'd just suggest the following change in analyze.ts:

Line 305 change:
```ts
const m = staticModules[specifier.startsWith('node:') ? specifier.slice(5) : specifier];
```

line 345 change:
```ts
const staticModule = staticModules[source.startsWith('node:') ? source.slice(5) : source];
```

ijjk (approved):
Thanks!

test/unit/asset-fs-inline-path-enc-es/input.js
```diff
@@ -1,4 +1,4 @@
-import fs from 'fs';
-import * as path from 'path';
+import fs from 'node:fs';
+import * as path from 'node:path';
```
styfle, line 2: Should this test be reverted so we cover both cases?

ijjk: Yeah we can revert this one since we have the separate fixtures now, updated in https://github.com/vercel/nft/pull/287

PR #287, opened 2022-05-18
chore: Update node prefix tests

ijjk:
Ensures we have the test with the `node:` prefix and without, also renames the `node` prefix fixtures to make it more clear that's what is being tested. 

x-ref: https://github.com/vercel/nft/pull/285#discussion_r876079458


styfle (approved):

test/unit/asset-fs-array-expr-node-require/output.js
```diff
@@ -1,5 +0,0 @@
-[
-  "package.json",
-  "test/unit/asset-fs-array-expr-node-require/asset1.txt",
-  "test/unit/asset-fs-array-expr-node-require/input.js"
-]
```
styfle, line 5: I was thinking the output.js file was necessary for each test fixture 🤔

ijjk: It's still necessary just was renamed https://github.com/vercel/nft/blob/main/test/unit/asset-fs-array-expr-node-prefix/output.js guess it shows as a delete because the content changed more than git expects.

styfle: Oh I see, git didn't track it as a rename since the contents changed too much 👍

PR #290, opened 2022-05-24
[dependencies] remove old node-pre-gyp

EndangeredMassa:
The `node-pre-gyp` package has been deprecated in favor of `@mapbox/node-pre-gyp`, which is already used in this project. This PR removes the last use of `node-pre-gyp`.

Removing this gets rid of the following warning that's shown when installing anything that depends on `vercel/nft`:

>  WARN  deprecated node-pre-gyp@0.13.0: Please upgrade to @mapbox/node-pre-gyp: the non-scoped node-pre-gyp package is deprecated and only the @mapbox scoped package will recieve updates in the future


EndangeredMassa:
@styfle Ahhh, the package name change included breaking changes. https://github.com/mapbox/node-pre-gyp/blob/master/CHANGELOG.md#100

It's not a huge deal to have a warning during package install. We could leave this as is for now. Or we could drop the unscoped package name entirely and release a new major version. What do you think?

ijjk (commented):
We could potentially ncc this package or copy over the relevant files to remove the warning and still have backwards compatibility

styfle (changes requested):
Let's add tests to verify these lines are executed

https://github.com/vercel/nft/blob/799fa3066a7ead0496d2bd5aeddbf9c1b7de234a/src/analyze.ts#L110-L112

EndangeredMassa:
I added a test, but it fails with:

```
should correctly trace and correctly execute node-pre-gyp.js

    expect(received).toBe(expected) // Object.is equality

    Expected: 0
    Received: 1

      51 |     });
      52 |     const code = await new Promise(resolve => ps.on('close', resolve));
    > 53 |     expect(code).toBe(fails ? 1 : 0);
         |                  ^
      54 |     rimraf.sync(tmpdir);
      55 |     
      56 |     // TODO: ensure analysis cache is safe for below case

      at Object.<anonymous> (test/integration.test.js:53:18)
```

It looks like `node-pre-gyp` never makes it into the traced file list, but I'm not sure why. If the reason it obvious to someone else, let me know. Otherwise I'll dig deeper.

EndangeredMassa:
I dug more into this, but it's not clear to me how to proceed. This issue isn't worth spending a lot more time on.

At some point I'd like to pair with someone who knows this repo well on this issue as a way for me to gain a better understanding of how the internals work.

PR #301, opened 2022-07-15
fix: add concurrency limit to fs calls

Brooooooklyn:
This adds concurrency limits to our `fs` calls to attempt to alleviate memory issues with too many promises being stored in memory at once. 

x-ref: [slack thread](https://vercel.slack.com/archives/CGU8HUTUH/p1657905624763969?thread_ts=1655847330.113019&cid=CGU8HUTUH)

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
   "dependencies": {
     "@mapbox/node-pre-gyp": "^1.0.5",
     "acorn": "^8.6.0",
+    "async-sema": "^3.1.1",
     "bindings": "^1.4.0",
     "estree-walker": "2.0.2",
     "glob": "^7.1.3",
--- a/src/node-file-trace.ts
+++ b/src/node-file-trace.ts
@@ -81,8 +81,8 @@ export class Job {
     ts = true,
     analysis = {},
     cache,
-    // 1024 may enough here, to prevent memory issues cause by too many concurrent file IO
-    // and do not hurt the performance much
+    // we use a default of 1024 concurrency to balance
+    // performance and memory usage for fs operations
     fileIOConcurrency = Number(process.env.FILE_IO_CONCURRENCY) || 1024,
   }: NodeFileTraceOptions) {
     this.ts = ts;
--- a/src/node-file-trace.ts
+++ b/src/node-file-trace.ts
@@ -83,7 +83,7 @@ export class Job {
     cache,
     // we use a default of 1024 concurrency to balance
     // performance and memory usage for fs operations
-    fileIOConcurrency = Number(process.env.FILE_IO_CONCURRENCY) || 1024,
+    fileIOConcurrency = 1024,
   }: NodeFileTraceOptions) {
     this.ts = ts;
     base = resolve(base);
@@ -161,7 +161,7 @@ export class Job {
     await this.fileIOQueue.acquire();
     try {
       const link = await fsReadlink(path);
-      await this.fileIOQueue.release();
+      this.fileIOQueue.release();
       // also copy stat cache to symlink
       const stats = this.statCache.get(path);
       if (stats)
@@ -170,7 +170,7 @@ export class Job {
       return link;
     }
     catch (e) {
-      await this.fileIOQueue.release();
+      this.fileIOQueue.release();
       if (e.code !== 'EINVAL' && e.code !== 'ENOENT' && e.code !== 'UNKNOWN')
         throw e;
       this.symlinkCache.set(path, null);
@@ -198,12 +198,12 @@ export class Job {
     await this.fileIOQueue.acquire();
     try {
       const stats = await fsStat(path);
-      await this.fileIOQueue.release();
+      this.fileIOQueue.release();
       this.statCache.set(path, stats);
       return stats;
     }
     catch (e) {
-      await this.fileIOQueue.release();
+      this.fileIOQueue.release();
       if (e.code === 'ENOENT') {
         this.statCache.set(path, null);
         return null;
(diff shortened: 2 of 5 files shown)


ijjk (approved):
Thanks!

package.json
```diff
@@ -48,6 +48,7 @@
     "analytics-node": "^3.4.0-beta.1",
     "apollo-server-express": "^2.14.2",
     "argon2": "^0.27.2",
+    "async-sema": "^3.1.1",
```
styfle, line 51: Is this used at runtime? It should be in `dependencies`, not `devDependencies`

src/node-file-trace.ts
```diff
@@ -79,6 +81,7 @@ export class Job {
     ts = true,
     analysis = {},
     cache,
+    fileIOConcurrency = 128,
```
styfle, line 84: Why 128? Can the user customize?

styfle (dismissed):
Please add a PR description. Also can you add a test to prove this fixes it? (presumably this is fixing something due to the title of the PR?)

src/node-file-trace.ts
```diff
@@ -116,6 +119,9 @@ export class Job {
     this.paths = resolvedPaths;
     this.log = log;
     this.mixedModules = mixedModules;
+    this.fileIOQueue = new Sema(fileIOConcurrency, {
+      capacity: fileIOConcurrency * 10,
```
styfle, line 123: Why `fileIOConcurrency * 10`?

src/node-file-trace.ts
```diff
@@ -152,8 +158,10 @@ export class Job {
   async readlink (path: string) {
     const cached = this.symlinkCache.get(path);
     if (cached !== undefined) return cached;
+    await this.fileIOQueue.acquire();
```
styfle, line 161: Any reason why this acquire is outside the `try` but the one on line 199 is inside the `try`?

src/node-file-trace.ts
ijjk: Good point, we probably want to remove the `capacity` here as we can't know ahead of time a good amount that will be needed.

src/node-file-trace.ts
Brooooooklyn: It's okay to `acquire` out of `try` block, I'll move the line `199` out of `try`

ijjk (approved):

src/node-file-trace.ts
```diff
@@ -152,8 +158,10 @@ export class Job {
...
     try {
       const link = await fsReadlink(path);
+      await this.fileIOQueue.release();
```
jridgewell, line 164: These [aren't](https://github.com/vercel/async-sema/blob/603842b82e68e24b76eb0e74dae66de1eda468b4/src/index.ts#L195-L197) async operations.

```suggestion
      this.fileIOQueue.release();
```

src/node-file-trace.ts
```diff
@@ -162,6 +170,7 @@ export class Job {
       return link;
     }
     catch (e) {
+      await this.fileIOQueue.release();
```
jridgewell, line 173:
```suggestion
      this.fileIOQueue.release();
```

src/node-file-trace.ts
```diff
@@ -186,12 +195,15 @@ export class Job {
   async stat (path: string) {
     const cached = this.statCache.get(path);
     if (cached) return cached;
+    await this.fileIOQueue.acquire();
     try {
       const stats = await fsStat(path);
+      await this.fileIOQueue.release();
```
jridgewell, line 201:
```suggestion
      this.fileIOQueue.release();
```

src/node-file-trace.ts
```diff
@@ -186,12 +195,15 @@ export class Job {
...
       this.statCache.set(path, stats);
       return stats;
     }
     catch (e) {
+      await this.fileIOQueue.release();
```
jridgewell, line 206:
```suggestion
      this.fileIOQueue.release();
```

src/node-file-trace.ts
```diff
@@ -207,12 +219,15 @@ export class Job {
   async readFile (path: string): Promise<string | Buffer | null> {
     const cached = this.fileCache.get(path);
     if (cached !== undefined) return cached;
+    await this.fileIOQueue.acquire();
     try {
       const source = (await fsReadFile(path)).toString();
       this.fileCache.set(path, source);
+      await this.fileIOQueue.release();
```
jridgewell, line 226: Nit: Flip these to make it consistent with the others:
```suggestion
      this.fileIOQueue.release();
      this.fileCache.set(path, source);
```

jridgewell (approved):

src/node-file-trace.ts
```diff
@@ -79,6 +81,9 @@ export class Job {
     ts = true,
     analysis = {},
     cache,
+    // we use a default of 1024 concurrency to balance
+    // performance and memory usage for fs operations
+    fileIOConcurrency = Number(process.env.FILE_IO_CONCURRENCY) || 1024,
```
styfle, line 86: Let not use an env var. Instead, the consumer should pass the `fileIOConcurrency` option to nodeFileTrace()

We also need to update the README.md with the new option

src/node-file-trace.ts
```diff
@@ -152,8 +158,10 @@ export class Job {
   async readlink (path: string) {
     const cached = this.symlinkCache.get(path);
     if (cached !== undefined) return cached;
+    await this.fileIOQueue.acquire();
     try {
       const link = await fsReadlink(path);
+      this.fileIOQueue.release();
```
styfle, line 164: I think we should move `release()` out of `try` and `catch` and instead move it into `finally` because I think there is a scenario where `release()` could be called twice and cause bugs.

Brooooooklyn: There is `return` in the try block, so we can't put the release in the `finally`. There is only the file op in the try block that may fail, so I think it would be okay that put release both in the try and catch block.

Brooooooklyn: TIL

styfle (approved):
Thanks!

ijjk (approved):

Issue #315, opened 2022-10-13
Bug with shadowed variable not being resolved correctly


dummdidumm:
Analyzing this code

```js
const path = 123;
let binding;
for (const path of ["asd"]) {
        binding = require(path);
}
exports.default = binding;
```

fails with the error
```
Failed to resolve dependency 123:
The "path" argument must be of type string. Received type number (123)
```

It seems that `path` is resolved to `123` and not to `"asd"`, and it seems that assigning to `binding` is what triggeres it, because if that is removed, it works.

The real world case where this happens is when analyzing `aws-crt`

styfle:
Thanks for reporting! 

Would you like to submit a PR with the fix?

dummdidumm:
I looked into it and I'm not sure how to proceed. The problem is the very outdated dependency `rollup-pluginutils` which now lives under `@rollup/pluginutils`. Using version 4 (version 5 requires node 14 at the minimum, which is probably too high for this package) of it correctly recognizes the for loop as a new scope. Using that prevents the analyze-code from crashing and therefore the warning from being emitted, but the dependency that is required isn't added to the list of files. I'm not sure if this is expected, or a different bug.

dummdidumm:
With #316 merged, is there anything left to do here? The dependency `"add"` isn't added, and I'm not sure if this is expected behavior or a missing case in static analysis.

styfle:
> version 5 requires node 14 at the minimum, which is probably too high for this package

We can a target node 14 now that 12 is EOL

iamabhshk:
To fix this code, you should avoid using the name path for both the loop variable and the constant variable at the beginning. Here's an updated version of the code:

```
const pathValue = 123;

let binding;
for (const p of ["asd"]) {
    binding = require(p);
}

exports.default = binding;

```

PR #316, opened 2022-10-14
fix: bump rollup pluginutils version

dummdidumm:
Part of #315

This fixes a bug where `attachScopes` didn't detect the new scope inside the for loop with shadowed variables before, resulting in later code paths possibly crashing because they expect the dependency to be of type string. No test added because as noted in #315 this doesn't thix that issue (if it even is one).

Bumping the version made a previous workaround of conflicting estree-walker versions unnecessary, but required to bring back a now-deleted type from the package

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   ],
   "dependencies": {
     "@mapbox/node-pre-gyp": "^1.0.5",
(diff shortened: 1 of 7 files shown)


styfle (approved):
Great, thanks! 🎉

PR #320, opened 2022-11-01
fix: Make caching work correctly in async context

danez:
Right now the cache was only populated after the call to the fs API succeed, which in an async context was leading to multiple file IO or analyze tasks started for the same file at once. This is because both file IO are async and so until the fs call finished other calls for the same file can still come in.

To fix this I extracted all file IO calls into a separate class that wraps everything correctly and instead of caching the actual result, now a Promise is cached. So the cache now always needs to be awaited.

As the values in the cache change this is definitely a breaking change. Something like this:

```diff
const cache = {};
await nodeFileTrace([file], {
    cache,
);

-const cacheValue = cache.fileCache.get(file);
+const cacheValue = await cache.fileCache.get(file);
```

I know this is quite a big change, just let me know what you think and if I should change anything. :) 

In my personal example the number of calls to analyze went from ~2400 to ~1600 and the runtime from ~3s to ~2.7s (which includes other work than nft)

### DIFF
--- a/src/fs.ts
+++ b/src/fs.ts
@@ -1,0 +1,116 @@
+import type { Stats } from "fs";
+import { resolve } from "path";
+import fs from "graceful-fs";
+import { Sema } from "async-sema";
+
+const fsReadFile = fs.promises.readFile;
+const fsReadlink = fs.promises.readlink;
+const fsStat = fs.promises.stat;
+
+export class CachedFileSystem {
+  private fileCache: Map<string, Promise<string | null>>;
+  private statCache: Map<string, Promise<Stats | null>>;
+  private symlinkCache: Map<string, Promise<string | null>>;
+  private fileIOQueue: Sema;
+
+  constructor({
+    cache,
+    fileIOConcurrency,
+  }: {
+    cache?: any;
+    fileIOConcurrency: number;
+  }) {
+    this.fileIOQueue = new Sema(fileIOConcurrency);
+    this.fileCache = (cache && cache.fileCache) || new Map();
+    this.statCache = (cache && cache.statCache) || new Map();
+    this.symlinkCache = (cache && cache.symlinkCache) || new Map();
+
+    if (cache) {
+      cache.fileCache = this.fileCache;
+      cache.statCache = this.statCache;
+      cache.symlinkCache = this.symlinkCache;
+    }
+  }
+
+  async readlink(path: string): Promise<string | null> {
+    const cached = this.symlinkCache.get(path);
+    if (cached !== undefined) return cached;
+    // This is not awaiting the response, so that the cache is instantly populated and
+    // future calls serve the Promise from the cache
+    const readlinkPromise = this.executeFileIO(path, this._internalReadlink);
+    this.symlinkCache.set(path, readlinkPromise);
+
+    return readlinkPromise;
+  }
+
+  async readFile(path: string): Promise<string | null> {
+    const cached = this.fileCache.get(path);
+    if (cached !== undefined) return cached;
+    // This is not awaiting the response, so that the cache is instantly populated and
+    // future calls serve the Promise from the cache
+    const readFilePromise = this.executeFileIO(path, this._internalReadFile);
+    this.fileCache.set(path, readFilePromise);
+
+    return readFilePromise;
+  }
+
+  async stat(path: string): Promise<Stats | null> {
+    const cached = this.statCache.get(path);
+    if (cached !== undefined) return cached;
+    // This is not awaiting the response, so that the cache is instantly populated and
+    // future calls serve the Promise from the cache
+    const statPromise = this.executeFileIO(path, this._internalStat);
+    this.statCache.set(path, statPromise);
+
+    return statPromise;
+  }
+
+  private async _internalReadlink(path: string) {
+    try {
+      const link = await fsReadlink(path);
+      // also copy stat cache to symlink
+      const stats = this.statCache.get(path);
+      if (stats) this.statCache.set(resolve(path, link), stats);
+      return link;
+    } catch (e: any) {
+      if (e.code !== "EINVAL" && e.code !== "ENOENT" && e.code !== "UNKNOWN")
+        throw e;
+      return null;
+    }
+  }
+
+  private async _internalReadFile(path: string): Promise<string | null> {
+    try {
+      return (await fsReadFile(path)).toString();
+    } catch (e: any) {
+      if (e.code === "ENOENT" || e.code === "EISDIR") {
+        return null;
+      }
+      throw e;
+    }
+  }
+
+  private async _internalStat(path: string) {
+    try {
+      return await fsStat(path);
+    } catch (e: any) {
+      if (e.code === "ENOENT") {
+        return null;
+      }
+      throw e;
+    }
+  }
+
+  private async executeFileIO<Return>(
+    path: string,
+    fileIO: (path: string) => Promise<Return>
+  ): Promise<Return> {
+    await this.fileIOQueue.acquire();
+
+    try {
+      return fileIO.call(this, path);
+    } finally {
+      this.fileIOQueue.release();
+    }
+  }
+}
--- a/src/fs.ts
+++ b/src/fs.ts
@@ -21,9 +21,9 @@ export class CachedFileSystem {
     fileIOConcurrency: number;
   }) {
     this.fileIOQueue = new Sema(fileIOConcurrency);
-    this.fileCache = (cache && cache.fileCache) || new Map();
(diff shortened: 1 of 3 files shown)


src/node-file-trace.ts
```diff
@@ -153,30 +160,29 @@ export class Job {
     this.esmFileList = new Set();
     this.processed = new Set();
     this.warnings = new Set();
+
+    // This creates the cached versions of these IO functions
+    // The internal methods are only there in case the functions are overwritten from config and the overwritten function wants
+    // to call the internal implementation
+    this.stat = this.internalStat = cacheFileIOFactory(this.statFn.bind(this), this.statCache, this.fileIOQueue);
+    this.readFile = this.internalReadFile = cacheFileIOFactory(this.readFileFn.bind(this), this.fileCache, this.fileIOQueue);
+    this.readlink = this.internalReadlink = cacheFileIOFactory(this.readlinkFn.bind(this), this.symlinkCache, this.fileIOQueue);
```
danez, line 169: internal might actually not be a good choice for the name, because it makes it look like something purely internal one shouldn't use. Maybe rather `base` or `real`. Naming :D

Maybe you think it is not even necessary to have these methods, then I can also remove them and we need to figure out a different way in the tests.

src/node-file-trace.ts
```diff
@@ -193,50 +199,32 @@ export class Job {
...
+  protected async statFn (path: string) {
     try {
-      const stats = await fsStat(path);
-      this.statCache.set(path, stats);
-      return stats;
+      return await fsStat(path);
     }
     catch (e: any) {
       if (e.code === 'ENOENT') {
-        this.statCache.set(path, null);
```
styfle, line 207: What about deleting from the cache?

The new wrapper doesn't do that 🤔

danez: The wrapper does cache the surrounding Promise, so in case of an error and when `null` is returned it will also be part if the cached promise.

As far as I understood how the cache works it doesn't do cache invalidation anyway. Once cached it is cached forever. :)

src/node-file-trace.ts
styfle: >  it makes it look like something purely internal one shouldn't use

I think it is internal and shouldn't be used. It was only added in PR #40 as a mock.

styfle (commented):
Is there a way to add a test to ensure this fixes the behavior?

danez:
I added tests, which fail on the main branch.
The tests ensure, that all the cached functions are called only once for each file.

On main:
Screenshot 2022-11-07 at 16 19 52

On this branch:
Screenshot 2022-11-07 at 16 20 07

test/unit.test.js
```diff
@@ -197,14 +235,34 @@ for (const { testName, isRoot } of unitTests) {
...
+        expect(stat).toHaveBeenCalledTimes(0);
+        expect(readlink).toHaveBeenCalledTimes(testName === 'processed-dependency' ? 1 : 0);
+        expect(readFile).toHaveBeenCalledTimes(testName === 'processed-dependency' ? 1 : 0);
+        expect(analyze).toHaveBeenCalledTimes(testName === 'processed-dependency' ? 1 : 0);
+      } else {
+        // Ensure all cached calls are only called once per file. The expected count is the count of calls unique per path
+        const uniqueStatCalls = [...new Set(stat.mock.calls.map((call) => call[0]))].length;
+        const uniqueReadlinkCalls = [...new Set(readlink.mock.calls.map((call) => call[0]))].length;
+        const uniqueReadFileCalls = [...new Set(readFile.mock.calls.map((call) => call[0]))].length;
+        const uniqueAnalyzeFileCalls = [...new Set(analyze.mock.calls.map((call) => call[0]))].length;
```
styfle, line 250: `Set` has a `.size` property you can use instead of `.length`

src/node-file-trace.ts
```diff
@@ -54,18 +55,24 @@ export class Job {
   public exportsOnly: boolean;
   public paths: Record<string, string>;
   public ignoreFn: (path: string, parent?: string) => boolean;
+  public stat: (path: string) => Promise<Stats | null>;
+  public readFile: (path: string) => Promise<Buffer | string | null>;
+  public readlink: (path: string) => Promise<string | null>;
+  protected internalStat: (path: string) => Promise<Stats | null>;
```
styfle, line 61: Lets change these to `private` and use `_` to prefix the variable name.

test/unit.test.js
danez: done

src/node-file-trace.ts
danez: done

src/node-file-trace.ts
```diff
@@ -54,18 +55,27 @@ export class Job {
...
   public ignoreFn: (path: string, parent?: string) => boolean;
+  public stat: (path: string) => Promise<Stats | null>;
+  public readFile: (path: string) => Promise<Buffer | string | null>;
+  public readlink: (path: string) => Promise<string | null>;
+  // @ts-expect-error not used inside class, but maybe in stat method from config
+  private _internalStat: (path: string) => Promise<Stats | null>;
+  // @ts-expect-error not used inside class, but maybe in readFile method from config
+  private _internalReadFile: (path: string) => Promise<Buffer | string | null>;
+  // @ts-expect-error not used inside class, but maybe in readlink method from config
+  private _internalReadlink: (path: string) => Promise<string | null>;
```
styfle, line 66: It looks like these underscore methods are assigned but never used

src/node-file-trace.ts
```diff
@@ -366,23 +366,19 @@ export class Job {
...
+      analyzeResultPromise = cachedAnalysis;
     }
     else {
-      const source = await this.readFile(path);
-      if (source === null) throw new Error('File ' + path + ' does not exist.');
-      // analyze should not have any side-effects e.g. calling `job.emitFile` 
-      // directly as this will not be included in the cachedAnalysis and won't
-      // be emit for successive runs that leverage the cache
-      analyzeResult = await analyze(path, source.toString(), this);
-      this.analysisCache.set(path, analyzeResult);
```
styfle, line 382: Lets move this code back to how it was and just remove the `await` (no need for a new runAnalyze function)

styfle (commented):
I tried to make some changes but it looks like I can't push to your branch so I left a few comments

src/fs.ts
```diff
@@ -0,0 +1,116 @@
...
+  private fileCache: Map<string, Promise<string | null>>;
+  private statCache: Map<string, Promise<Stats | null>>;
+  private symlinkCache: Map<string, Promise<string | null>>;
+  private fileIOQueue: Sema;
+
+  constructor({
+    cache,
+    fileIOConcurrency,
+  }: {
+    cache?: any;
```
styfle, line 20:
```suggestion
    cache?: { fileCache?: Map<string, Promise<string | null>>, statCache?: Map<string, Promise<Stats | null>>, symlinkCache?: Map<string, Promise<string | null>> };
```

src/fs.ts
```diff
@@ -0,0 +1,116 @@
...
+    fileIOConcurrency: number;
+  }) {
+    this.fileIOQueue = new Sema(fileIOConcurrency);
+    this.fileCache = (cache && cache.fileCache) || new Map();
+    this.statCache = (cache && cache.statCache) || new Map();
+    this.symlinkCache = (cache && cache.symlinkCache) || new Map();
```
styfle, line 26:
```suggestion
    this.fileCache = cache?.fileCache ?? new Map();
    this.statCache = cache?.statCache ?? new Map();
    this.symlinkCache = cache?.symlinkCache ?? new Map();
```

src/node-file-trace.ts
```diff
@@ -137,15 +128,9 @@ export class Job {
       }, analysis === true ? {} : analysis);
     }
 
-    this.fileCache = cache && cache.fileCache || new Map();
-    this.statCache = cache && cache.statCache || new Map();
-    this.symlinkCache = cache && cache.symlinkCache || new Map();
     this.analysisCache = cache && cache.analysisCache || new Map();
```
styfle, line 131: Should analysisCache also go in the `CachedFileSystem` class?

danez: It is not used there and right now it is only used in the job class.

If we move the cache we would need to move the analysis functionality there too, which I'm not sure fits into `CachedFileSystem `

styfle (approved):
Great work, thanks 🎉

danez:
It seems because the fork I create is inside an organization it does not work, GitHub does not allow edits from maintainers.  😒https://github.com/orgs/community/discussions/5634

I merged the main branch, hopefully that allows merging :)

PR #336, opened 2022-12-21
fix(resolution): Fix resolution of querystringed imports in ESM files

### LINKED ISSUE #319: Resolving files with querystring
Hi, I am one of the maintainers of the [`@sentry/nextjs` SDK](https://www.npmjs.com/package/@sentry/nextjs) and we recently got a [bug report](https://github.com/getsentry/sentry-javascript/issues/5998) stating that we mess up the output of `@vercel/nft`. This is completely valid and reproducible.

What we're doing in the SDK is that we have a webpack loader that "proxies" user files and re-exports wrapped versions of particular functions (i.e. Next.js data fetching methods and API routes).

For the proxying mechanism we first replace the user-file with a proxy-file, from which we then import the user-file with a query string at the end to mark it as "already proxied" to avoid infinite loops (e.g. `import * as wrapee from './page.js?__sentry_wrapped__'`). The [rollup docs](https://rollupjs.org/guide/en/#big-list-of-options:~:text=You%20can%20use%20this%20for%20instance%20as%20a%20mechanism%20to%20define%20custom%20proxy%20modules%20for%20entry%20points) explain this rather well.

For some reason this breaks the Next.js' nft file generation, leading to my first question: Do you know why this would be happening?

Webpack is able to resolve imports containing a query parameters - it simply strips the query parameter when looking for the file in the file system. Is nft also able to resolve these paths? I tried using `nft why` to experiment but it seemed like it couldn't resolve it with the error:

```text
Error: Failed to resolve dependency ./pages/index.js?__sentry_wrapped__:
Cannot find module '/workspaces/sentry-nextjs-vercel-testproject/pages/index.js?__sentry_wrapped__' loaded from /workspaces/sentry-nextjs-vercel-testproject/test.js
```

Do you know if there is a workaround for us or something we can do here? We would also be ready to contribute upstream to support query params and fragments in import paths.

lforst:
Note: This PR is built on top of https://github.com/vercel/nft/pull/322 which I took over because @lobsterkatie is currently out. 

---

Currently, there are two problems with the way `@vercel/nft` handles ESM imports with querystrings (`import * as something from './someFile?someQuerystring';`):

* It treats the `?` as it would any other character. As a result, it fails to resolve such imports to the non-querystringed files to which they correspond. (In the above example, it looks for a file named `someFile?someQuerystring.js` and, as one would expect, doesn't find it.) While this is the correct behavior for `require` calls in CJS files, it doesn't match the way that Node handles ESM imports (which is to strip the querystring off before attempting to resolve the file).
* In cases where a custom `resolve` method is provided, and that custom method _does_ strip querystrings from ESM imports, the querystrings are then not put back on before the imported module is processed. This matters, for example, in cases where `readFile` is also overridden, with a method whose output is affected by the querystring's value. (Webpack is such a system, so one place this shows up is in [Next.js's `NextTraceEntryPointsPlugin`](https://github.com/vercel/next.js/blob/3d4c038fa5c314f5800b8caeee7ada8c385cb3ce/packages/next/build/webpack/plugins/next-trace-entrypoints-plugin.ts).)

This fixes both of those problems by selectively stripping and re-adding the querystring in ESM imports at various points in the `nodeFileTrace` lifecycle. More specifically, in `resolveDependecy` and in `emitDependency` and its helpers, we strip the querystring (and/or don't add it back), with the exception of:

* When we're marking a path  as seen, we keep the querystring so that having a different querystring makes the process run again.
* When we're calling `readFile`, we keep the querystring, since some implementations of `readFile`, including the one in the nft webpack plugin, read different modules out of memory with different querystrings).
* When we're calling `resolve`, we add the querystring back in, since this is also something which can be supplied by the user, and and the user-supplied version might care about query strings.
* When we're caching analysis results, we keep the querystring, since again here, we may have different code to analyze if we have a different querystring.
* When we're making the recursive call to `emitDependency`, we add the querystring back in, since we need it there to be able to start the whole cycle over.

Notes:

* Because the behavior here depends on whether or not a module is ESM, I needed to pass `cjsResolve` to the recursive call of `emitDependency`. Rather than change `emitDependency`'s signature, I opted to make `emitDependency` a wrapper for an inner function (`analyzeAndEmitDependency`) with the updated signature. Recursive calls now go to the inner function.
* For context, we on the `@sentry/nextjs` team are interested in this because we use a [webpack loader](https://github.com/getsentry/sentry-javascript/blob/c9a01ff98cdde31de217f0d47ad2626f867d6eaa/packages/nextjs/src/config/loaders/proxyLoader.ts) to insert in the dependency tree a [proxy module](https://github.com/getsentry/sentry-javascript/blob/c9a01ff98cdde31de217f0d47ad2626f867d6eaa/packages/nextjs/src/config/templates/pageProxyLoaderTemplate.ts) for each page, which allows us to automatically wrap users' exported functions (`getServerSideProps` and friends, API route handlers, etc). To do this, we replace the code from `/path/to/pages/somePage.js` with our proxy code, which includes an import from `'./somePage?__sentry_wrapped__'`. When webpack then crawls our proxy module looking for dependencies, the querystring tricks it into thinking it hasn't yet dealt with that page, so it forwards it to our loader a second time rather than skipping it. When our loader sees the querystring, it knows the code replacement has already been done, and returns the original page code untouched, thereby preserving the continuity of the dependency tree. This fix ensures that `@vercel/nft` successfully traces the modified tree.
* The unit test testing that this doesn't break the (already correct) behavior in CJS when dealing with files with question marks in their names (`cjs-querystring`) required a little gymnastics in order to not break CI on Windows. It turns out that skipping the test on Windows wasn't enough - the mere fact that a file exists with a `?` in its name is enough to cause Windows not to even be able to check out the commit. Therefore, the file is stored without any punctuation in its name, and before the tests run on non-Windows platforms, it is copied to a version which does have the punctuation, and which is git-ignored to enforce it not ending up in the repo in the future by accident. The dummy, no-punctuation version is stored in a folder in order to differentiate its path from that of the punctuated version by more than just the punctuation. That way, if the punctuated version is found, it's clear it's not just because the punctuation is somehow being ignored.

Fixes [getsentry/sentry-javascript#5998](https://github.com/getsentry/sentry-javascript/issues/5998).
Fixes #319.

### DIFF
--- a/src/node-file-trace.ts
+++ b/src/node-file-trace.ts
@@ -266,16 +266,12 @@ export class Job {
       return;
     }
 
-    if (Array.isArray(resolved)) {
-      for (const item of resolved) {
-        // ignore builtins
-        if (item.startsWith('node:')) return;
-        await this.analyzeAndEmitDependency(item, path, cjsResolve);
-      }
-    } else {
+    // For simplicity, force `resolved` to be an array
+    resolved = Array.isArray(resolved) ? resolved : [resolved];
+    for (const item of resolved) {
       // ignore builtins
-      if (resolved.startsWith('node:')) return;
-      await this.analyzeAndEmitDependency(resolved, path, cjsResolve);
+      if (item.startsWith('node:')) return;
+      await this.analyzeAndEmitDependency(item, path, cjsResolve);
     }
   }
 
--- a/src/node-file-trace.ts
+++ b/src/node-file-trace.ts
@@ -375,13 +375,15 @@ export class Job {
     // Strip the querystring, if any. (Only affects ESM dependencies.)
     const { path } = parseSpecifier(rawPath, cjsResolve)
 
-    if (this.processed.has(path)) {
+    // Since different querystrings may lead to different results, include the full path
+    // when noting whether or not we've already seen this path
+    if (this.processed.has(rawPath)) {
       if (parent) {
         await this.emitFile(path, 'dependency', parent)
       }
       return
     };
-    this.processed.add(path);
+    this.processed.add(rawPath);
 
     const emitted = await this.emitFile(path, 'dependency', parent);
     if (!emitted) return;
@@ -400,7 +402,8 @@ export class Job {
 
     let analyzeResult: AnalyzeResult;
 
-    const cachedAnalysis = this.analysisCache.get(path);
+    // Since different querystrings may lead to analyses, include the full path when caching
+    const cachedAnalysis = this.analysisCache.get(rawPath);
     if (cachedAnalysis) {
       analyzeResult = cachedAnalysis;
     }
@@ -411,7 +414,7 @@ export class Job {
       // directly as this will not be included in the cachedAnalysis and won't
       // be emit for successive runs that leverage the cache
       analyzeResult = await analyze(path, source.toString(), this);
-      this.analysisCache.set(path, analyzeResult);
+      this.analysisCache.set(rawPath, analyzeResult);
     }
 
     const { deps, imports, assets, isESM } = analyzeResult;
--- a/src/node-file-trace.ts
+++ b/src/node-file-trace.ts
@@ -248,12 +248,12 @@ export class Job {
     } catch (e1: any) {
       error = e1;
       try {
-        if (this.ts && dep.endsWith('.js') && e1 instanceof NotFoundError) {
+        if (this.ts && strippedDep.endsWith('.js') && e1 instanceof NotFoundError) {
           // TS with ESM relative import paths need full extensions
           // (we have to write import "./foo.js" instead of import "./foo")
           // See https://www.typescriptlang.org/docs/handbook/esm-node.html
-          const depTS = dep.slice(0, -3) + '.ts';
-          resolved = await this.resolve(depTS, path, this, cjsResolve);
+          const strippedDepTS = strippedDep.slice(0, -3) + '.ts';
+          resolved = await this.resolve(strippedDepTS + queryString, path, this, cjsResolve);
           error = undefined;
         }
       } catch (e2: any) {
--- a/src/node-file-trace.ts
+++ b/src/node-file-trace.ts
@@ -370,7 +370,11 @@ export class Job {
     return this.analyzeAndEmitDependency(path, parent)
   }
 
-  private async analyzeAndEmitDependency(path: string, parent?: string, cjsResolve?: boolean) {
+  private async analyzeAndEmitDependency(rawPath: string, parent?: string, cjsResolve?: boolean) {
+
+    // Strip the querystring, if any. (Only affects ESM dependencies.)
+    const { path } = parseSpecifier(rawPath, cjsResolve)
+
     if (this.processed.has(path)) {
       if (parent) {
         await this.emitFile(path, 'dependency', parent)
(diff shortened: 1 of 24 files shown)


lforst:
@styfle New PR because @lobsterkatie is out over the holidays. You mind taking a look? It's the same changes but with the feedback on `url.parse` you wrote.

Btw, feel free to push any changes you want to make directly to this PR!

styfle:
@lforst There are merge conflicts in the original PR as well as this PR

src/node-file-trace.ts
```diff
@@ -1,17 +1,42 @@
...
+export function parseSpecifier(specifier: string, cjsResolve: boolean = true): ParseSpecifierResult {
+  let path = specifier;
+  let queryString = null;
+
+  if (!cjsResolve) {
+    const specifierUrl = url.parse(specifier);
+    queryString = specifierUrl.search;
+
+    if (specifierUrl.search) {
+      path = specifier.replace(specifierUrl.search, '');
```
styfle, line 31: This could replace the wrong part of the URL if it had repeating content in the string. 

How about using the parsed result instead?

```suggestion
      path = specifierUrl.pathname;
```

src/node-file-trace.ts
```diff
@@ -1,17 +1,42 @@
...
+}
+
+// Splits an ESM specifier into path and querystring (including the leading `?`). (If the specifier is CJS,
+// it is passed through untouched.)
+export function parseSpecifier(specifier: string, cjsResolve: boolean = true): ParseSpecifierResult {
+  let path = specifier;
+  let queryString = null;
+
+  if (!cjsResolve) {
+    const specifierUrl = url.parse(specifier);
```
styfle, line 27: I just learned `url.parse()` was deprecated (again) so lets use `new URL()` instead

- https://github.com/nodejs/node/pull/44919

lforst: Any suggestion on how to deal with URL not being able to handle relative paths? Would it be ok for us to use some random base url (eg `example.com`)?

src/node-file-trace.ts
lforst: Because we want to exclusively strip the query string off the `specifier`. If we use `pathname` we would also strip away suff like hash, protocol, host etc. (I know it is a bit edge-casey but still, let's do as little as possible to the specifier here)

I will rename the function to something like `splitQueryStringFromSpecifier` and change the return object signature to `{ queryString, remainingSpecifier }` to make its purpose more clear.

lforst: As for the repeating content, do you have examples of where this might go wrong?

Afaik, as per URL standard, the first `?` will denote the start of the query/`search` part of the URL. `search` should always start with a `?` and since `.replace()` will only replace the first occurrence of a match (unless something like the regex `g` flag is used) this should just work IMO.

styfle: > we would also strip away stuff like hash, protocol, host etc.

Did you add tests for those cases? How does Node.js handle those cases?

>  since .replace() will only replace the first occurrence of a match

Ah yes, you're right.

> new URL("../../src/mymodule", "file://").pathname === "/src/mymodule")

Oh good catch!

src/node-file-trace.ts
styfle: I think this is a job for [`pathToFileURL()`](https://nodejs.org/api/url.html#urlpathtofileurlpath)

lforst:
We don't need this anymore

PR #354, opened 2023-07-13
feat(resolve): export `resolve()` function

gnoff:
nft allows you to override the resolve step with a hook however it does not expose its internal resolveDependency implementation. This means if you want to take over resolving in certain cases you must do so in every case. To make conditional or augmented resolving possible this change reexports the resolveDependency function from the index. This way you can use it in custom resolve hook implementations

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,4 @@
 export * from './types';
 export { nodeFileTrace } from './node-file-trace';
 import resolveDependency from './resolve-dependency';
-export { resolveDependency };
+export { resolveDependency as resolve };
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,2 +1,4 @@
 export * from './types';
 export { nodeFileTrace } from './node-file-trace';
+import resolveDependency from './resolve-dependency';
+export { resolveDependency };
--- a/readme.md
+++ b/readme.md
@@ -109,6 +109,25 @@ The following FS functions can be hooked by passing them as options:
 * `readlink(path): Promise<string>`
 * `resolve(id: string, parent: string): Promise<string | string[]>`
 
+##### Advanced Resolving
+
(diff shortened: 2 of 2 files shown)


src/index.ts
```diff
@@ -1,2 +1,4 @@
 export * from './types';
-export { nodeFileTrace } from './node-file-trace';
\ No newline at end of file
+export { nodeFileTrace } from './node-file-trace';
+import resolveDependency from './resolve-dependency';
+export { resolveDependency };
```
styfle, line 4:
```suggestion
export { resolveDependency };

```

styfle (commented):
Sounds good!

Lets also document it in the README since this is the default `resolve()` hook behavior.

I wonder if we should export as `resolve` so it matches the hook? Do you have a preference?

```ts
import { nodeFileTrace , resolve } from '@vercel/nft' 

nodeFileTrace([file.txt], { resolve })
```

gnoff:
> Lets also document it in the README since this is the default resolve() hook behavior.

Hmm this API leaves a little to be desired. there are 2 arguments to `resolveDependency` that aren't really public, the `job` and `isCjs` arguments.

Perhaps we ought to provide a `resolve` function to the `resolve` hook which curries these arguments to avoid exposing them. It's somewhat unfortunate that the extra arguments are already provided when they are not documented because we can't really change them now. I'll think a bit more about this

gnoff:
I renamed the export and described it as it is currently implemented in the readme but still don't love the API

ijjk (approved):

gnoff:
@styfle is there something up with ci?

styfle:
@gnoff Yep, fixed in https://github.com/vercel/nft/pull/355

styfle (approved):
Thanks!

styfle:
> we can't really change them now

We can change make breaking changes if we ship under a new version.

We haven't even hit 1.0.0 so breaking changes are expected https://github.com/vercel/nft/releases

That being said, I think we should ship what we have to unblock the work in Next.js and we can make changes if needed in a new PR in the future.

gnoff:
Yeah sounds good!

PR #358, opened 2023-08-09
fix: use `builtinModules` from `module`

danielroe:
This feels like the better choice rather than using a private property on `repl`.

supported as of [v6.13.0](https://nodejs.org/api/module.html#modulebuiltinmodules)


styfle (approved):
Thanks!

ijjk (approved):

PR #361, opened 2023-09-09
fix: support pattern trailers

jedwards1211:
Supports trailing characters after the `*` in wildcard `exports` patterns, e.g.

```
    "./*.js": {
      "module": "./*.js",
      "default": {
        "node": "./*b.js"
      }
    },
```

I copied the logic from [Node's resolve implementation](https://github.com/nodejs/node/blob/main/lib/internal/modules/esm/resolve.js#L539C1-L539C1).

### DIFF
--- a/src/resolve-dependency.ts
+++ b/src/resolve-dependency.ts
@@ -274,7 +274,8 @@ function resolveExportsImports(
       const patternIndex = target.indexOf("*");
       if (patternIndex !== -1 && target.lastIndexOf("*") === patternIndex) {
         return (
-          target.substring(0, patternIndex) +
+          pkgPath +
+          target.substring(1, patternIndex) +
           bestMatchSubpath +
           target.substring(patternIndex + 1)
         );
--- a/src/resolve-dependency.ts
+++ b/src/resolve-dependency.ts
@@ -1,84 +1,127 @@
-import { isAbsolute, resolve, sep } from 'path';
-import { builtinModules } from 'module';
-import { Job } from './node-file-trace';
+import { isAbsolute, resolve, sep } from "path";
+import { builtinModules } from "module";
+import { Job } from "./node-file-trace";
 
 // node resolver
 // custom implementation to emit only needed package.json files for resolver
 // (package.json files are emitted as they are hit)
-export default async function resolveDependency (specifier: string, parent: string, job: Job, cjsResolve = true): Promise<string | string[]> {
+export default async function resolveDependency(
+  specifier: string,
+  parent: string,
+  job: Job,
+  cjsResolve = true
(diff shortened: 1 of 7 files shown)


src/resolve-dependency.ts
```diff
@@ -284,60 +214,33 @@ function resolveExportsImports(
     }
   }
 
-  for (const match of Object.keys(matchObj).sort(
-    (a, b) => b.length - a.length
-  )) {
-    if (!match.endsWith("/")) continue;
+  for (const match of Object.keys(matchObj).sort((a, b) => b.length - a.length)) {
```
guybedford, line 217: This sort should be updated to use the order defined by `PATTERN_KEY_COMPARE` in https://nodejs.org/dist/latest-v20.x/docs/api/esm.html#resolution-and-loading-algorithm.

src/resolve-dependency.ts
```diff
@@ -1,127 +1,84 @@
...
-    path.startsWith(job.base) &&
-    path.slice(job.base.length).indexOf(sep + "node_modules" + sep) === -1 &&
-    (await job.isFile(path + ".tsx"))
-  )
-    return path + ".tsx";
-  if (await job.isFile(path + ".js")) return path + ".js";
-  if (await job.isFile(path + ".json")) return path + ".json";
-  if (await job.isFile(path + ".node")) return path + ".node";
+  if (job.ts && path.startsWith(job.base) && path.slice(job.base.length).indexOf(sep + 'node_modules' + sep) === -1 && await job.isFile(path + '.ts')) return path + '.ts';
+  if (job.ts && path.startsWith(job.base) && path.slice(job.base.length).indexOf(sep + 'node_modules' + sep) === -1 && await job.isFile(path + '.tsx')) return path + '.tsx';
```
guybedford, line 43: Perhaps abstract this into an `extensions` configuration rather, so it can be configured and not necessarily the default (sincethis will match `./file.ts` before `file.js`).

guybedford (commented):
Thanks for posting this, it is definitely a missing feature and the last missing feature before this resolver implementation can be considered complete (given this resolve feature landed in Node.js later itself).

I left a couple of comments. That said I don't think the implementation is complete and likely needs a closer look at the spec. It also needs tests to land.

If you need implementation inspiration, perhaps take a look at https://github.com/rollup/plugins/pull/1549 which was very nicely written.

jedwards1211:
Okay I'll go over this again soon.

Side note: it seems like a shame to have all these separate implementations of the node module resolution algorithm.  Is there really no package that's flexible enough to be used in all of these cases?  I've used the `resolve` package some before, but I haven't investigated if it's lacking anything or why we wouldn't want to depend on it here.

PR #362, opened 2023-09-15
fix: resolve cjs deps as cjs instead of esm

cb1kenobi:
When `nft` analyzes a file that has both a `require()` and an ESM `export`, the first call to acorn fails because of the `export`, so it treats the whole file as ESM, which is probably fine.

The problem is it treats the `require()` as an ESM import which is wrong. This will cause the resolve dependency entrypoint in the dependency to get the ESM export instead of the proper CJS export.

If the dependency is an `import`, add it to `imports`. If it's a `require()`, add it to `deps` and don't worry about whatever `isESM`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "@types/graceful-fs": "^4.1.5",
     "@types/micromatch": "^4.0.1",
     "@types/node": "^14.14.37",
+    "@vercel/edge-config": "^0.4.0",
     "analytics-node": "^3.4.0-beta.1",
     "apollo-server-express": "^2.14.2",
     "argon2": "^0.27.2",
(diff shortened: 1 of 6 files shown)


styfle (changes requested):
Please add a test

package.json
```diff
@@ -45,6 +45,7 @@
     "@types/graceful-fs": "^4.1.5",
     "@types/micromatch": "^4.0.1",
     "@types/node": "^14.14.37",
+    "@vercel/edge-config": "^0.4.0",
```
styfle, line 48: This needs to be added to the lock file too. You might need to run in linux x64 to update since some of the dependencies don't work on arm unfortunately.

cb1kenobi: Done

styfle (approved):
Looks good if tests pass, thanks!

styfle (approved):
Great, thanks!

PR #364, opened 2023-10-02
demonstrate mjs file identified as esm, but handled as cjs

cb1kenobi:


styfle:
@cb1kenobi This test is passing https://github.com/vercel/nft/actions/runs/6384547761/job/17327429573?pr=364#step:8:6107

Is that because it doesn't set `mixedModules: true`?

cb1kenobi:
@styfle Correct. I didn't commit the change to `integration.test.js` that enables `mixedModules = true` for this test.

PR #365, opened 2023-10-02
fix: Revert "fix: resolve cjs deps as cjs instead of esm (#362)"

cb1kenobi:
This reverts commit 1eba4d507bcfc0d34ae51d98ed1b285dc4537923.

The fix in the original PR did solve the problem of tracing a dependency being loaded from an ESM file via `require()`, but also caused a regression where a simple `.mjs` file with ESM imports was failing. I have a test (https://github.com/vercel/nft/pull/364) for that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "@types/graceful-fs": "^4.1.5",
     "@types/micromatch": "^4.0.1",
     "@types/node": "^14.14.37",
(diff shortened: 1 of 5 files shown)


styfle (approved):

styfle:
@cb1kenobi Can you update the PR description to explain why its reverted?

PR #366, opened 2023-10-11
fix: add shiki to special case list

souporserious:
Adds shiki to the special case list so grammars and themes will be traced and loaded correctly when deploying.


test/integration/shiki.js
```diff
@@ -0,0 +1,3 @@
+const javascriptGrammar = require('shiki/languages/javascript.tmLanguage.json');
```
styfle, line 1: This test should use shiki as you intend to use it, not as the workaround

test/unit/shiki/input.js
```diff
@@ -0,0 +1,3 @@
+const javascriptGrammar = require('shiki/languages/javascript.tmLanguage.json');
```
styfle, line 1: Similarly here, it should be the normal usage.

Also you need to add an output file for unit tests to compare the traced files include the language.

souporserious: Oops didn't realize this should go in output. Updated!

souporserious:
@styfle thank you for the feedback! I updated the input/output files to reflect the real-world usage. However, for some reason I'm still having trouble with tests not passing.

styfle:
@souporserious I pushed a change which should fix it

souporserious:
@styfle woot, that looks great, thank you! 🙏  I can't tell why the test is still failing, but I'm wondering if we need to account for themes as well since I believe `shiki` will look for the theme in the example use since it's specifying `nord`.

styfle (approved):
I pushed a couple more changes. If tests pass, we should be good, thanks! ✅

Issue #371, opened 2023-11-30
`@vercel/nft` does not pick up platform-specific binaries for `sharp`


devunt:
Related:
- https://github.com/lovell/sharp/issues/3870
- https://github.com/getsentry/sentry-cli/issues/1849

`@vercel/nft` does not pick up platform-specific binary files while tracing dependencies.

For example, `sharp@0.33.0` [depends on](https://github.com/lovell/sharp/blob/main/lib/sharp.js#L16) `@img/sharp-{arch}` package, but the traced result only contains below paths.

```
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/package.json
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/lib/index.js
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/lib/constructor.js
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/lib/input.js
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/lib/resize.js
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/lib/composite.js
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/lib/colour.js
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/lib/channel.js
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/lib/utility.js
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/lib/operation.js
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/lib/output.js
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/lib/is.js
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/lib/sharp.js
node_modules/.pnpm/sharp@0.33.0/node_modules/sharp/lib/libvips.js
node_modules/.pnpm/sharp@0.33.0/node_modules/color
node_modules/.pnpm/sharp@0.33.0/node_modules/detect-libc
node_modules/.pnpm/sharp@0.33.0/node_modules/semver
```

This currently breaks every build on Vercel and any deployments that uses `@vercel/nft` as its tree shaking library.

lovell:
The logic used for `esbuild` should solve this:

https://github.com/vercel/nft/blob/37811197975fef8a3cda7d17e436d217a510398c/src/utils/special-cases.ts#L32-L41

[...]

PR #372, opened 2023-12-01
fix: add special case for sharp@0.33.0 optional dependencies

styfle:
- Fixes https://github.com/vercel/nft/issues/371

Additionally:
- Replaced `yarn` with `npm` (sharp requires it)
- Updated `@ffmpeg-installer/ffmpeg`, `oracledb`, `sharp` to versions which support Apple Silicon
- Disabled `microtime-node-gyp` test on Apple Silicon
- Fixed expected output for some tests which implicitly upgraded dep versions

I also skipped a few ECMAScript Next tests because they currently fail.

At first I thought this must be a regression since the tests were passing on main (head is 285c9309ee041599d31202bac3bad6bbb05aba22). However, the tests were always failing and jest was silently succeeding. Regenerating the lockfile, implicitly upgraded jest from `27.4.5` to `27.5.1` (due to the `^` prefix) and this included a fix that is now surfacing the error.

I also converted this file from `jest` to `node:test` just to confirm that it also fails on main.

Since these tests are Stage 2 features, likely no one has noticed they don't work because no JS runtime implements them.


styfle:
Looks like `sharp` doesn't support `yarn@1` anymore since it didn't install `optionalDependencies` 🤔 

https://github.com/vercel/nft/actions/runs/7054667887/job/19203908329?pr=372#step:8:7545

So we might need to change the repo to use npm or at least this specific test needs npm.

devunt:
Adding `--ignore-engines` on `yarn install` per https://github.com/lovell/sharp/issues/3871 might solve the installation issue.

lovell:
yarn v1 has been [in maintenance mode for almost 4 years](https://dev.to/arcanis/introducing-yarn-2-4eh1#what-will-happen-to-the-legacy-codebase) and does not properly support `optionalDependencies`. I highly recommend people upgrade, but please avoid the plug'n'play feature.

cb1kenobi:
@styfle Whatcha think about this PR? Merge it?

.gitignore
```diff
@@ -4,3 +4,6 @@ out
 coverage
 test/**/dist
 test/**/actual.js
+package-lock.json
+yarn.lock
+pnpm-lock.yaml
```
styfle, line 9: @cb1kenobi  I think we should keep one lockfile here so that we have deterministic behavior.

test/ecmascript/data-esnext.js
styfle, line 1: @cb1kenobi This looks like a regression if the existing tests are failing

.gitignore
cb1kenobi: Fixed!

.github/workflows/ci.yml
```diff
@@ -41,13 +41,13 @@ jobs:
     - name: Prepare Install (if applicable)
       run: node prepare-install.js
     - name: Install Dependencies
-      run: yarn install --network-timeout 1000000
+      run: npm install --no-package-lock
```
styfle, line 44:
```suggestion
      run: npm install
```

prepare-install.js
```diff
@@ -7,7 +7,6 @@ if (isWin) {
   const pkgJson = readFileSync(join(__dirname, 'package.json'), 'utf8');
   const pkg = JSON.parse(pkgJson);
 
-  //unlinkSync(join(__dirname, 'yarn.lock'));
   // Delete the integration tests that will never work in Windows
```
styfle, line 10:
```suggestion
  unlinkSync(join(__dirname, 'package-lock.json'));
  // Delete the integration tests that will never work in Windows
```

test/ecmascript/data-esnext.js
cb1kenobi: Yeah, probably. I tinkered with it for a bit and I'm those tests are for proposed ECMAScript features that still haven't been accepted. If you wanna take a peek at them, then be my guest, but I'm all for moving onward and upward.

lovell:
Great to see some progress on this, thank you.

The sharp-related change adds 10 lines whereas the majority of this PR appears to relate to replacing yarn v1 with the latest npm. If clean version control history is of interest then may I suggest a split into a PR for the package manager upgrade and a PR to support `optionalDependencies`.

styfle:
Even if we split it into 2 PRs, it will still have the same problem because we can't run tests without installing `sharp`. 

This should be wrapping up soon though, there was a bug in jest causing some confusion and I think we have it resolved.

test/ecmascript.test.js
```diff
@@ -13,6 +13,9 @@ console.log('created directory ' + tmpdir);
 
  // These are tests known to fail so we skip them
 const ignoreCategories = new Set([
+  'Generator function.sent Meta Property',
+  'Class and Property Decorators',
+  'throw expressions',
 ]);
```
styfle, line 19: We are skipping these tests now because they currently fail.

At first I thought this must be a regression since the tests were passing on main (head is 285c9309ee041599d31202bac3bad6bbb05aba22). However, the tests were always failing and jest was silently succeeding. Regenerating the lockfile, implicitly upgraded jest from `27.4.5` to `27.5.1` (due to the `^` prefix) and this included a fix that is now surfacing the error.

I also converted this file from `jest` to `node:test` just to confirm that it also fails on main.

Since these tests are Stage 2 features, likely no one has noticed they don't work because no JS runtime implements them.

styfle:
Looks like we're running into this npm bug now 🤦 

- https://github.com/npm/cli/issues/4828

cb1kenobi (approved):

PR #377, opened 2023-12-11
feat: support import attributes

Skn0tt:
The [Import Attributes proposal](https://github.com/tc39/proposal-import-attributes) is in Stage 3, so implementation is started. Currently, NFT bails on any file that includes import attributes, because Acorn can't parse them. That's because Acorn only supports Stage 4 proposals - but for import attributes, there's a plugin that this PR adds to NFT.


Skn0tt:
It seems like `acorn-import-attributes` doesn't support the `assert { type: "json" }` syntax, even though that's still part of the proposal. I'll try to contribute a fix for that directly to `acorn-import-attributes`.

cb1kenobi (approved):
LGTM

PR #387, opened 2024-01-31
fix: update sharp special case for pnpm

ijjk:
This ensures we trace child optional dependencies as well so that the symlink hierarchy is maintained for pnpm.   

x-ref: https://github.com/vercel/next.js/issues/59346
x-ref: https://github.com/lovell/sharp/issues/3967


styfle (approved):
I believe we're running into a windows bug on node@18 so I had to skip it https://github.com/nodejs/node/issues/18518

But all other platforms are passing, thanks! 🎉

PR #392, opened 2024-02-13
fix(analyze): node-gyp-build not including zeromq prebuilds

### LINKED ISSUE #391: Add support for `zeromq`
Some libraries like `zeromq.js` need runtime libraries with bindings which they ship along. Those binaries exist under `node_modules/zeromq/prebuilds`.

Simply by creating a new Next.js app, installing https://github.com/zeromq/zeromq.js and trying to import it and running `npm run dev` and use one of its classes would throw a `No native build was found for platform=linux arch=x64 runtime=node abi=115 uv=1 libc=glibc node=20.11.0 webpack=true loaded from: /app/.next/server`. I found out that if I print [this `dir` variable](https://github.com/aminya/node-gyp-build/blob/4cd1eb0f1ef2ed7d4c36b82e3d547e33cd83c9a0/index.js#L26) used (among other things) to search for binaries in node-gyp-builder it print `.next/server/...` instead of `node_modules/...` and since the binaries have not been copied it fails.

> [!NOTE]
> I provided a reproducible example and a container, but it's easily reproducible on mac and linux: https://github.com/eulersson/zeromq.js-next.js-errors

I tried using the `outputFileTracingIncludes` option as explained in https://nextjs.org/docs/pages/api-reference/next-config-js/output#caveats but without success.

Here's the files I would like to have in `.next/server`.

```
❯ tree node_modules/zeromq/prebuilds
node_modules/zeromq/prebuilds
├── darwin-arm64
│   └── node.napi.glibc.node
├── darwin-x64
│   └── node.napi.glibc.node
├── linux-x64
│   ├── node.napi.glibc.node
│   └── node.napi.musl.node
├── win32-ia32
│   └── node.napi.glibc.node
└── win32-x64
    └── node.napi.glibc.node

6 directories, 6 files
```

Since I don't know in what area work should be done I posted the same problem as an issue to the various tools involved:

- https://github.com/vercel/next.js/issues/61844
- https://github.com/zeromq/zeromq.js/issues/600
- https://github.com/prebuild/node-gyp-build/issues/65

Thanks for the help and apologies if I have been generating noise by posting so many issues.

eulersson:
- Fixes #391 

Three things I want to point out:

1. In zeromq the way node-gyp-build is called is by passing `path.join(__dirname, "..")` instead of `__dirname` to the `require("node-gyp-build")(...)` call.
2. In zeromq they use `@aminya/node-gyp-build` instead of `node-gyp-build`, particular-li version `4.5.0-aminya.5`.
3. The zeromq version that's recommended for new project is the `6` which is beta according to @aminya https://github.com/zeromq/zeromq.js.

After those changes it resolves OK.

```
❯ node ./node_modules/@vercel/nft/out/cli.js print ./node_modules/zeromq/lib/index.js
FILELIST:
node_modules/@aminya/node-gyp-build/index.js
node_modules/@aminya/node-gyp-build/package.json
node_modules/zeromq/lib/draft.js
node_modules/zeromq/lib/index.js
node_modules/zeromq/lib/native.js
node_modules/zeromq/lib/util.js
node_modules/zeromq/package.json
node_modules/zeromq/prebuilds/darwin-x64/node.napi.glibc.node <---- It worked!

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,7 +130,7 @@
     "vue": "^2.6.10",
     "vue-server-renderer": "^2.6.10",
     "when": "^3.7.8",
-    "zeromq": "6.0.0-beta.19"
+    "zeromq": "^6.0.0-beta.19"
   },
   "engines": {
     "node": ">=16"
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "build": "tsc",
     "prepublishOnly": "tsc && rm out/utils/*.d.ts && rm out/tsconfig.tsbuildinfo",
     "test": "jest --verbose",
+    "test-watch": "jest --watch",
     "test-verbose": "tsc --sourceMap && jest --verbose --coverage --globals \"{\\\"coverage\\\":true}\""
   },
   "files": [
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
     "out"
   ],
   "dependencies": {
-    "@aminya/node-gyp-build": "4.5.0-aminya.5",
     "@mapbox/node-pre-gyp": "^1.0.5",
     "@rollup/pluginutils": "^4.0.0",
     "acorn": "^8.6.0",
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "out"
   ],
   "dependencies": {
+    "@aminya/node-gyp-build": "4.5.0-aminya.5",
     "@mapbox/node-pre-gyp": "^1.0.5",
     "@rollup/pluginutils": "^4.0.0",
     "acorn": "^8.6.0",
--- a/package.json
+++ b/package.json
@@ -129,7 +129,8 @@
     "vm2": "^3.9.18",
     "vue": "^2.6.10",
     "vue-server-renderer": "^2.6.10",
-    "when": "^3.7.8"
+    "when": "^3.7.8",
+    "zeromq": "6.0.0-beta.19"
   },
   "engines": {
     "node": ">=16"
--- a/src/analyze.ts
+++ b/src/analyze.ts
@@ -15,6 +15,8 @@ import resolve from './resolve-dependency.js';
 //@ts-ignore
 import nodeGypBuild from 'node-gyp-build';
 //@ts-ignore
+import aminyaNodeGypBuild from '@aminya/node-gyp-build';
+//@ts-ignore
 import mapboxPregyp from '@mapbox/node-pre-gyp';
 import { Job } from './node-file-trace';
 import { fileURLToPath, pathToFileURL, URL } from 'url';
@@ -125,6 +127,9 @@ const staticModules = Object.assign(Object.create(null), {
   'node-pre-gyp/lib/pre-binding': pregyp,
   'node-pre-gyp/lib/pre-binding.js': pregyp,
   'node-gyp-build': {
+    default: NODE_GYP_BUILD
+  },
+  '@aminya/node-gyp-build': {
     default: NODE_GYP_BUILD
   },
   'nbind': {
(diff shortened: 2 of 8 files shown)


test/unit.test.js
```diff
@@ -218,6 +218,21 @@ for (const { testName, isRoot } of unitTests) {
         expect(foundMatchingBinary).toBe(true)
       }
 
+      if (testName === 'zeromq-node-gyp') {
```
styfle, line 221: Maybe we should move this into line 206 so the logic is the same for both `microtime-node-gyp` and `zeromq-node-gyp`

test/unit/zeromq-node-gyp/input.js
```diff
@@ -0,0 +1 @@
+require('zeromq');
```
styfle, line 1: A unit test is great! Let's also add an integration test. Basically copy this same file to `test/integration/zeromq.js`

eulersson:
I will fix the test havoc ⛈️

test/unit/zeromq-node-gyp/input.js
eulersson: Done.

test/unit.test.js
eulersson: Done.

test/unit/zeromq-node-gyp/output.js
```diff
@@ -1,9 +1,11 @@
 [
+  "node_modules/@aminya/node-gyp-build/index.js",
+  "node_modules/@aminya/node-gyp-build/package.json",
   "node_modules/zeromq/lib/draft.js",
   "node_modules/zeromq/lib/index.js",
   "node_modules/zeromq/lib/native.js",
   "node_modules/zeromq/lib/util.js",
   "node_modules/zeromq/package.json",
-  "node_modules/zeromq/prebuilds/darwin-x64/node.napi.glibc.node",
```
eulersson: @styfle  I thought during the tests the prebuilds would get picked up but they don't. I have seen the PR you gave me as example the output list also did not contain the binary: https://github.com/vercel/nft/pull/286/files#diff-fd366a1b2d1df15ab6dcbf1c7ee76e755a9ce56ee74e247e5657738061e65c9e (test/unit/microtime-node-gyp/output.js), so I am assuming it's OK, since if I install the package locally on a Next.js app it works fine.

package.json
```diff
@@ -12,12 +12,14 @@
     "build": "tsc",
     "prepublishOnly": "tsc && rm out/utils/*.d.ts && rm out/tsconfig.tsbuildinfo",
     "test": "jest --verbose",
+    "test-watch": "jest --watch",
```
styfle, line 15:
```suggestion
```

src/analyze.ts
```diff
@@ -614,18 +619,57 @@ export default async function analyze(id: string, code: string, job: Job): Promi
...
+
+                // Resolve the path in the case of passing path.join(__dirname, "..") instead of __dirname.
+                const pathJoinedDir =
+                  calledWithPathJoinDirname ? path.join(dir, node.arguments[0].arguments[1].value) : dir;
+
+                // zeromq.js uses @aminya's fork of node-gyp-build, consider it.
+                const validNodeGypBuildForkNames = ['node-gyp-build', '@aminya/node-gyp-build'];
+                if (node.callee.arguments[0].type !== 'Literal' ||
+                    !validNodeGypBuildForkNames.includes(node.callee.arguments[0].value)) {
+                  throw new Error("Unknown node-gyp-build fork name");
```
styfle, line 649: We shouldn't throw. This will be impossible for users to comprehend.

src/analyze.ts
```diff
@@ -614,18 +619,57 @@ export default async function analyze(id: string, code: string, job: Job): Promi
...
-                    resolved = nodeGypBuild.path(dir);
+                    switch (nodeGypBuildForkName) {
+                      case "node-gyp-build":
+                        resolved = nodeGypBuild.path(pathJoinedDir);
+                        break;
+                      case "@aminya/node-gyp-build":
+                        resolved = aminyaNodeGypBuild.path(pathJoinedDir);
+                        break;
+                      default:
+                        throw new Error("Unknown node-gyp-build fork name");
```
styfle, line 671: Same here: we shouldn't throw

src/analyze.ts
eulersson: What would you say it's best then? Console error?

src/analyze.ts
styfle, line 671:
```suggestion
                        job.warnings.addnew Error(`Unknown node-gyp-build fork name: ${nodeGypBuildForkName`'));
```

src/analyze.ts
styfle: Skip here and add a warning there: https://github.com/vercel/nft/pull/392#discussion_r1488419981

aminya:
@eulersson Thanks for making a fix. Let me know if there is anything to do on the Zeromq side.

src/analyze.ts
eulersson: I will do! :) I don't think it would get there anyway for now unless there's drifting or disynchronization between what has been defined as `NODE_GYP_BUILD` symbol and what exists in the list `validNodeGypBuildForkNames`. In the future it could happen... Or a developer that adds a new symbol that resolves to `NODE_GYP_BUILD` but does not type it in the aforementioned array... Good to have I guess, then.

eulersson: Done.

src/analyze.ts
eulersson: Done.

eulersson:
> @eulersson Thanks for making a fix. Let me know if there is anything to do on the Zeromq side.

No problem. For now I do not think it will be needed.

styfle (approved):
Great work, thanks! 🎉

PR #417, opened 2024-05-09
feat: support import assertions

onsclom:
`acorn-import-attributes` [did not support import assertions before](https://github.com/vercel/nft/pull/377#issuecomment-1849719137), but now it does!

This is important because `acorn` is failing to parse `runner.js` in [this issue](https://github.com/vercel/vercel/issues/11533) due to an import assertion.


TooTallNate (approved):

trek (approved):

styfle (approved):
Thanks!

styfle (commented):
@onsclom I believe `assert` was renamed to `with`.

Can you also add a test for `with` as well?

https://github.com/tc39/proposal-import-attributes

onsclom:
> @onsclom I believe `assert` was renamed to `with`.
> 
> Can you also add a test for `with` as well?
> 
> https://github.com/tc39/proposal-import-attributes

[This PR](https://github.com/vercel/nft/pull/377/files) introduced a test for `with` so we are good there!

PR #422, opened 2024-06-03
fix: concat performance

styfle:
work in progress

### DIFF
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -129,6 +129,7 @@ for (const { testName, isRoot } of unitTests) {
         inputFileNames.push('input-2.js', 'input-3.js', 'input-4.js');
       }
 
+      const startTime = Date.now();
       const { fileList, reasons } = await nodeFileTrace(
         inputFileNames.map((file) => join(unitPath, file)),
         {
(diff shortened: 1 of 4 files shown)


src/utils/safe-stringify.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+  if (typeof value === 'function') {
+    return '[Function]';
+  }
+  if (typeof value === 'bigint') {
+    return '[bigint]';
+  }
+  return value;
+};
+export function safeStringify(obj: unknown) {
```
styfle, line 32: We need a more comprehensive stringify to handle all possible inputs

src/utils/static-eval.ts
```diff
@@ -1,29 +1,38 @@
+import { safeStringify } from './safe-stringify';
 import { EvaluatedValue, StaticValue, ConditionalValue, Node } from './types';
-import { URL } from 'url';
-type Walk = (node: Node) => EvaluatedValue;
+type Walk = (node: Node) => Promise<EvaluatedValue>;
 type State = { computeBranches: boolean; vars: Record<string, any> };
 
+const walkCache = new Map<string, EvaluatedValue>();
```
styfle, line 6: We should probably use a LRU cache to avoid running out of memory. Right now, the number of entries is unbounded.

ojj1123:
Is this PR also related with https://github.com/vercel/next.js/issues/65512 ?

PR #424, opened 2024-06-04
fix: Support browser overrides in package.json

dvoytenko:
`browser` section `package.json` is used to expand a list of dependent files. The downstream build tool would normally do the right thing and use the `browser` section for bundling correctly. I tested the bundling itself on esbuild and WebPack, and it works correctly. All that's needed is to make sure the files are there.


ijjk (approved):
This makes sense to me and will be good so we can trace for react-server and other conditions as well

PR #429, opened 2024-06-25
fix: Support resolving `module.register` dependencies

### LINKED ISSUE #428: Does not include registered ESM loader hooks
`package.json`
```json
{
  "type": "module",
  "dependencies": {
    "import-in-the-middle": "^1.8.1"
  },
  "devDependencies": {
    "@vercel/nft": "^0.27.2"
  }
}
```
`test.js`
```ts
import { register } from 'module';

register('import-in-the-middle/hook.mjs', import.meta.url);
```
`build.js`
```ts
import { nodeFileTrace } from '@vercel/nft';
const { fileList } = await nodeFileTrace(['./test.js']);
console.log(fileList)
```

Running `> node build.js`
Outputs `Set(1) { 'test.js' }`

But `import-in-the-middle/hook.mjs` and all of it's dependencies should be included.

timfish:
Closes #428

### DIFF
--- a/src/analyze.ts
+++ b/src/analyze.ts
@@ -67,6 +67,7 @@ const FS_FN = Symbol();
 const FS_DIR_FN = Symbol();
 const BINDINGS = Symbol();
 const NODE_GYP_BUILD = Symbol();
+const MODULE_FN = Symbol();
 const fsSymbols = {
   access: FS_FN,
   accessSync: FS_FN,
@@ -94,6 +95,9 @@ const fsExtraSymbols = {
   readJsonSync: FS_FN,
   readJSONSync: FS_FN,
 };
+const moduleSymbols = {
+  register: MODULE_FN,
+};
 const staticModules = Object.assign(Object.create(null), {
   bindings: {
     default: BINDINGS,
@@ -110,6 +114,10 @@ const staticModules = Object.assign(Object.create(null), {
   fs: {
     default: fsSymbols,
     ...fsSymbols,
+  },
+  module: {
+    default: moduleSymbols,
+    ...moduleSymbols,
   },
   'fs-extra': {
     default: fsExtraSymbols,
@@ -875,6 +883,14 @@ export default async function analyze(
               )
                 pjsonPath = path.resolve(pjsonPath, '../../package.json');
               if (pjsonPath !== rootPjson) assets.add(pjsonPath);
+              break;
+            case MODULE_FN:
+              if (
+                node.arguments.length &&
+                node.arguments[0].type === 'Literal'
+              ) {
+                await processRequireArg(node.arguments[0]);
+              }
               break;
           }
         }
--- a/package-lock.json
+++ b/package-lock.json
@@ -71,6 +71,7 @@
         "graphql": "^14.4.2",
         "highlights": "^3.1.6",
         "hot-shots": "^6.3.0",
+        "import-in-the-middle": "^1.8.1",
         "ioredis": "^4.11.1",
         "isomorphic-unfetch": "^3.0.0",
         "jest": "^27.4.5",
@@ -15761,6 +15762,18 @@
         "node": ">=4"
       }
     },
+    "node_modules/import-in-the-middle": {
+      "version": "1.8.1",
+      "resolved": "https://registry.npmjs.org/import-in-the-middle/-/import-in-the-middle-1.8.1.tgz",
+      "integrity": "sha512-yhRwoHtiLGvmSozNOALgjRPFI6uYsds60EoMqqnXyyv+JOIW/BrrLejuTGBt+bq0T5tLzOHrN0T7xYTm4Qt/ng==",
+      "dev": true,
+      "dependencies": {
+        "acorn": "^8.8.2",
+        "acorn-import-attributes": "^1.9.5",
+        "cjs-module-lexer": "^1.2.2",
+        "module-details-from-path": "^1.0.3"
+      }
+    },
     "node_modules/import-lazy": {
       "version": "2.1.0",
       "resolved": "https://registry.npmjs.org/import-lazy/-/import-lazy-2.1.0.tgz",
@@ -20800,6 +20813,12 @@
       "dependencies": {
         "safe-buffer": "~5.1.0"
       }
+    },
+    "node_modules/module-details-from-path": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/module-details-from-path/-/module-details-from-path-1.0.3.tgz",
+      "integrity": "sha512-ySViT69/76t8VhE1xXHK6Ch4NcDd26gx0MzKXLO+F7NOtnqH68d9zF94nT8ZWSxXh8ELOERsnJO/sWt1xZYw5A==",
+      "dev": true
     },
     "node_modules/moment": {
       "version": "2.30.1",
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "graphql": "^14.4.2",
     "highlights": "^3.1.6",
     "hot-shots": "^6.3.0",
+    "import-in-the-middle": "^1.8.1",
     "ioredis": "^4.11.1",
     "isomorphic-unfetch": "^3.0.0",
     "jest": "^27.4.5",
--- a/test/integration/module-register.mjs
+++ b/test/integration/module-register.mjs
@@ -1,0 +1,3 @@
+import { register } from 'module';
+
+register('import-in-the-middle/hook.mjs', import.meta.url);


styfle (commented):
Thanks for the PR!

Can you change the integration test into a unit test instead? 

That way we can see see the input and output to make sure these remain the same instead of relying on implementation details of a 3rd party package.

src/analyze.ts
```diff
@@ -876,6 +884,14 @@ export default async function analyze(
                 pjsonPath = path.resolve(pjsonPath, '../../package.json');
               if (pjsonPath !== rootPjson) assets.add(pjsonPath);
               break;
+            case MODULE_FN:
+              if (
+                node.arguments.length &&
+                node.arguments[0].type === 'Literal'
+              ) {
+                await processRequireArg(node.arguments[0]);
```
styfle, line 892: It looks like `module.register(specifier[, parentURL][, options])` has more than one argument.

Do we need to handle these as well?

styfle: Also it should be handled as an import, not require:

> Customization hooks to be registered; this should be the same string that would be passed to import(), except that if it is relative, it is resolved relative to parentURL.

https://nodejs.org/api/module.html#moduleregisterspecifier-parenturl-options

Something like this:

https://github.com/timfish/nft/blob/db7751dc417f0de81d5b76141ea2e392deb3ecd2/src/analyze.ts#L647-L648

timfish: `parentURL` is usually always `import.meta.url` in ESM. Do we need to cater for other `parentURL`s, or can we safely assume its relative to the current file?

styfle: It could be any URL, correct? 

A quick search on google shows its not always `import.meta.url`


https://github.com/search?type=code&q=%22+module.register%28%22+language%3ATypeScript

https://github.com/dai-shi/waku/blob/aa8b021a2f1ab073abc601129b5d0d6c3bda8ca1/packages/waku/src/lib/renderers/dev-worker-impl.ts#L38

timfish: `parentURL` could potentially be any URL but `import.meta.url` and `pathToFileURL('./')` both essentially mean:

> if `specifier` is a relative path, resolve relative to the current file/directory

Code could potentially pass a different URL here but I suspect that is not common. When I'm back online after this flight I'll have a more thorough search!

styfle: I think we need to be careful here because we might trace the wrong file if we don't consider that parent url.

I believe we support `require.resolve()` and that can be relative to a path, as well as `await import()` although not sure if that one can be relative to a path.

timfish: It's probably enough to only resolve  `module.register` when we can statically analyse the parentURL trivially and only the simple/common cases matter. We can skip resolving everything else and log a warning or something.

`require.resolve()` does not behave exactly the same as the default ESM resolver so we should not be using that for resolving ESM. And as far as I know, node doesn't expose its default ESM resolver either which makes this painful. 

For example, `require.resolve()` automatically tries multiple file extensions but the ESM resolver expects file extensions to be included.

timfish: Actually there's `import.meta.resolve()` from node v20.6.0 but I think we can only access that from ESM 😂

styfle: I think its fine if we don't support all possible cases - meaning if we can't statically analyze then we don't emit.

The thing I'm not okay with is emitting the wrong files because this is recursive so walking down the wrong tree can lead to severe bloat.

So perhaps starting with these two common cases:

- `import.meta.url`
- `pathToFileURL('./')`

Then add a TODO comment mentioning that more complex values are not handled at this time.

test/unit/module-register/register.mjs
```diff
@@ -0,0 +1,5 @@
+import { register } from 'module';
+import { pathToFileURL } from 'url';
+
+register('./hook.mjs', import.meta.url);
+register('./hook2.mjs', pathToFileURL(__filename));
```
styfle, line 5: How about a test for `pathToFileURL('./')`?

test/unit/module-register/input.js
```diff
@@ -0,0 +1 @@
+require('./register.mjs')
```
styfle, line 1: Should this be `import('./register.mjs')` instead since this is presumably cjs?

Ideally you should be able to run `node test/unit/module-register/input.js` and it should work.

timfish: Ah, I just looked at some of the other tests and assumed `input.js` was always CJS so did this to load an ESM test file

test/unit/module-register/register.mjs
timfish: I also want to add a test for the bare specifier code path too

test/unit/module-register/input.js
timfish: Yes, this code in `register.mjs` cannot be in `input.js` because that is expected to be CJS. We need ESM to test `import.meta.url`.

We need this code to jump into ESM mode
```ts
require('./register.mjs')
```

src/analyze.ts
```diff
@@ -876,6 +893,46 @@ export default async function analyze(
...
+                    const parentURL =
+                      computedParentURL.value instanceof URL
+                        ? computedParentURL.value.href
+                        : computedParentURL.value;
+                    // Resolve the srcURL from the parentURL
+                    const srcURL = new URL(pathOrSpecifier, parentURL).href;
+
+                    // Resolve the srcPath relative to the current file
+                    const srcPath = path.relative(
+                      path.dirname(importMetaUrl),
```
styfle, line 920: Is it safe to call `dirname()` on a url. I thought this can only be used on paths?

timfish: A file URL isn't much different to a Windows absolute path with a drive letter so I guess it just works?

Do the tests pass on Windows?

Is there a better alternative?

timfish: I've removed this `dirname` usage!

timfish:
Looks like I need to pull my Windows laptop out and work out why those tests are failing!

timfish:
So the failing unit test (`zeromq-node-gyp`) passes in isolation on my Windows machine but fails when all the tests are run.

One thing I don't understand is how it's passing at all, even on other platforms.

The expected output is:
https://github.com/vercel/nft/blob/31ab41706201cadf09763273199af2aebd6dfcc2/test/unit/zeromq-node-gyp/output.js#L1-L11

But it's failing with an extra file included:
```
        "node_modules\\@aminya\\node-gyp-build\\index.js",
    +   "node_modules\\@aminya\\node-gyp-build\\node-gyp-build.js",
        "node_modules\\@aminya\\node-gyp-build\\package.json",
        "node_modules\\zeromq\\lib\\draft.js",
        "node_modules\\zeromq\\lib\\index.js",
        "node_modules\\zeromq\\lib\\native.js",
        "node_modules\\zeromq\\lib\\util.js",
```
And the reason it's being included:
```
      'node_modules\\@aminya\\node-gyp-build\\node-gyp-build.js' => {
        type: [ 'dependency' ],
        ignored: false,
        parents: Set(1) { 'node_modules\\@aminya\\node-gyp-build\\index.js' }
      }
```
And then when you look at [the actual source of the parent](https://github.com/aminya/node-gyp-build/blob/fb5ec8454147f5e7d69a8592292ac8724a554576/index.js#L5), it's obvious why this is being included:
```ts
} else { // else use the runtime version here
  module.exports = require('./node-gyp-build.js')
}
```
So how is this passing on every other platform but failing on Windows when the full test suite is run?

styfle:
This might fix it: https://github.com/vercel/nft/pull/432

I merged it in to see if tests pass now 🤞

src/analyze.ts
```diff
@@ -876,6 +893,49 @@ export default async function analyze(
...
+                      0,
+                      importMetaUrl.lastIndexOf('/'),
+                    );
+
+                    // Resolve the srcPath relative to the current file
+                    const srcPath = path.relative(currentDirURL, srcURL);
+                    // Make it a proper relative path
+                    const relativeSrcPath = srcPath.startsWith('.')
+                      ? srcPath
+                      : './' + srcPath;
```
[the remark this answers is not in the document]
styfle: Is this supposed to be a file path or a url import?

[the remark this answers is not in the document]
timfish: Fixed!

timfish:
Unfortunatly not 😭

I'm more confused as to how this test passes on any platform. The additional dependency in the failing case looks like it should be included!

styfle:
Looks like the only failing test is the new one now

```
 ● should correctly trace module-register from cwd

    expect(received).toEqual(expected) // deep equality

    - Expected  - 3
    + Received  + 0

      Array [
        "package.json",
    -   "test\\unit\\module-register\\hook.mjs",
    -   "test\\unit\\module-register\\hook2.mjs",
    -   "test\\unit\\module-register\\hook3.mjs",
        "test\\unit\\module-register\\input-esm.mjs",
        "test\\unit\\module-register\\input.js",
        "test\\unit\\module-register\\node_modules\\test-pkg\\index.mjs",
        "test\\unit\\module-register\\node_modules\\test-pkg\\package.json",
      ]
```

https://github.com/vercel/nft/actions/runs/9785332243/job/27018410588?pr=429#step:9:1063

timfish:
On my Windows 11 machine the `module-register` test passes.

image

I probably need to revisit this at a more reasonable hour in the morning!

timfish:
So I reverted the `path.sep` change and the tests now all pass on Windows.

Looking at the resulting `deps` and `imports` and on Windows for ALL the tests, these all contain forwards slashes `/`. It appears that Node `fs` and `path` are perfectly happy dealing with these on Windows.

s1gr1d:
@styfle can this be merged?

src/analyze.ts
```diff
@@ -876,6 +893,49 @@ export default async function analyze(
...
+                node.arguments[0].type === 'Literal'
+              ) {
+                const pathOrSpecifier = node.arguments[0].value;
+                // It's a relative URL
+                if (pathOrSpecifier.startsWith('.')) {
+                  // Compute the parentURL if it's statically analyzable
+                  const computedParentURL =
+                    node.arguments.length > 1
+                      ? await computePureStaticValue(node.arguments[1])
+                      : undefined;
```
EndangeredMassa, line 909: **suggestion (blocking):** ‏consider all signatures

The second argument can be `parentURL` or `options`, but this code assumes it's `parentURL`. Can we verify the type, then if it is `options`, pluck off `options.parentURL`?

Docs: https://nodejs.org/api/module.html#moduleregisterspecifier-parenturl-options

EndangeredMassa (changes requested):
Looks good, but there's one case we should add.

Thanks!

src/analyze.ts
onsclom: Can confirm, this is fixed and the test case was added!

onsclom (approved):

styfle (approved):
LGTM, thanks!

@EndangeredMassa Want to give it another look?

EndangeredMassa (approved):

Issue #421, opened 2024-05-23
Deprecated subdependencies found in @mapbox/node-pre-gyp


cdwmhcc:
## Description

When installing dependencies for our project that uses `@vercel/nft`, we encountered warnings about deprecated subdependencies. These warnings are caused by the `@mapbox/node-pre-gyp` package, which is a dependency of `@vercel/nft`.

## Details

The following deprecated subdependencies are reported during `pnpm install`:

- `are-we-there-yet@2.0.0`
- `gauge@3.0.2`
- `inflight@1.0.6`
- `npmlog@5.0.1`

## Expected Behavior

Dependencies within `@vercel/nft` should not rely on deprecated packages to ensure better security and compatibility with modern development environments.

## Additional Information

- Node.js version: [20.13.1]
- pnpm version: [9.1.2]
- Operating System: [windows 11]

styfle:
There is some related discussion here: https://github.com/vercel/nft/pull/407#issuecomment-2108830341

Would you like to submit a PR to fix it?

benmccann:
>I wonder if we even need to include the dependency at all. I looked at argon2 to see how its used and its just a single function export:

Compare https://npmgraph.js.org/?q=argon2 vs https://npmgraph.js.org/?q=argon2@0.31.1

benmccann:
I sent https://github.com/vercel/nft/pull/431 to upgrade `argon2`. I'm not sure what else is required to address this issue, but I believe this new version of argon2 should hopefully address the blockers and make this issue much easier to solve

benmccann:
I see. That makes much more sense to me. Thanks!

It's not exactly clear to me that `find` is easily replaceable though it would be awesome if someone can figure out how to do that. In the meantime I'll work on cleaning up `@mapbox/node-pre-gyp`. I'll track that here:

- https://github.com/mapbox/node-pre-gyp/pull/721
- https://github.com/mapbox/node-pre-gyp/pull/721

benmccann:
`argon2` switched from `@mapbox/node-pre-gyp` to `node-gyp-build` here:
https://github.com/ranisalt/node-argon2/commit/b47602840a259946039db8526ddd182d1430f634#diff-bc704b883867dea430073059d0e3061c8ac87c037826b06007646f667777916d

The latter library is dramatically lighter. I wonder if we could make the same switch here

benmccann:
@styfle based on the link from my past comment, do you think something like this might work?

[...]

PR #431, opened 2024-06-26
chore: upgrade argon2

benmccann:
ref https://github.com/vercel/nft/issues/421. I think this should unblock that issue, but I believe there is more required. Would you be able to fix it as a follow up or at least provide some pointers?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,7 +45,7 @@
         "@vercel/style-guide": "^5.2.0",
         "analytics-node": "^3.4.0-beta.1",
         "apollo-server-express": "^2.14.2",
-        "argon2": "^0.31.1",
+        "argon2": "^0.40.3",
         "auth0": "^3.0.1",
         "axios": "^1.6.0",
         "azure-storage": "^2.10.3",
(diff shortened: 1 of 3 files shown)


styfle (commented):
Since `argon2` is a devDependency, it's not related to https://github.com/vercel/nft/issues/421

That said, its fine to upgrade but it looks like this is causing code coverage to drop which makes me think its no longer testing the code path for argon in special-cases.ts

https://github.com/vercel/nft/blob/bd68351519b29e72a4be2e870893515c8cc0456a/src/utils/special-cases.ts#L21-L26

And more importantly, a the `zeromq-node-gyp` test started failing

https://github.com/vercel/nft/actions/runs/9686185024/job/26738120405?pr=431#step:10:1098

benmccann:
> Since argon2 is a devDependency, it's not related to https://github.com/vercel/nft/issues/421

Could you clarify your comment from https://github.com/vercel/nft/pull/407#issuecomment-2113765135? I thought you were saying the only reason `@mapbox/node-pre-gyp` is in `dependencies` is because `argon2` uses it. The old version of `argon2` depends on it, but the new version doesn't, so it sounded to me like upgrading `argon2` would allow us to remove `@mapbox/node-pre-gyp` thus fixing https://github.com/vercel/nft/issues/421

styfle:
@benmccann My comment was about replacing `@mapbox/node-pre-gyp` with our own implementation since its only using `find()`. I was using argon2 as an example of one of the packages that uses `@mapbox/node-pre-gyp` since I've never used it before so I had to look at how its used in real packages.

benmccann:
Ah, I see. Thanks for clarifying. I'll go ahead and close this then since it won't help with my goal of resolving https://github.com/vercel/nft/issues/421

PR #432, opened 2024-07-02
chore(test): Remove windows-specific `prepare-install`

jeffsee55:
There's a need to skip some integration tests for Windows, this PR moves the logic for skipping to the test setup itself rather than altering the tests and dependencies as we were doing in the `prepare-install` step. The pre-install logic was also deleting the `package-lock.json`, I'm not 100% clear on what the purpose was, but seems like it may have been done accidentally given the context of the [diff here](https://github.com/vercel/nft/pull/372#discussion_r1439789439).

With the previous logic, CI started to fail on Windows because the `package-lock.json` was removed before install, causing us to install a later version of dependency. Instead of removing that line of code in the `prepare-install` file, just trying this approach which feels a little more straightforward.


test/integration.test.js
```diff
@@ -12,7 +12,28 @@ jest.setTimeout(200_000);
...
+    ].includes(testName)
+  ) {
+    return false;
+  }
+  return true;
+});
+
+for (const integrationTest of filteredTestsToRun) {
+  console.log('Integration test: ' + integrationTest);
+
```
TooTallNate, line 36: Remove?

```suggestion
```

TooTallNate (approved):

styfle (approved):

PR #446, opened 2024-10-22
fix: add special case for `geoip-lite`

kongmoumou:
Currently, nft can not handle data assets of `geoip-lite` package properly. Error message like below ⬇️

image

Test Code

```js
const geoip = require('geoip-lite');

const ip = "207.97.227.239";
const geo = geoip.lookup(ip);
```

This PR adds essential data files as assets for `geoip-lite` in special-case to solve it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -68,6 +68,7 @@
         "firebase-admin": "^12.0.0",
         "fluent-ffmpeg": "^2.1.2",
         "geo-tz": "^7.0.1",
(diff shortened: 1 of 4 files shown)


styfle (approved):
Looks good!

We can merge this one but I think a better solution would be fixing the static analysis for the fallback here:

https://github.com/geoip-lite/node-geoip/blob/c788342dfce6d97244842622da6e4fa259f79742/lib/geoip.js#L13-L16

PR #452, opened 2024-11-26
fix: bump @rollup/plugin-utils to 5.1.3

benmccann:
Allows it to dedupe with the version most people would have in their projects if they're using rollup


styfle (approved):
Looks like the only breaking change from 4 to 5 was dropping node@14 support so we should be good since tests are passing, thanks!

PR #458, opened 2024-12-18
fix: avoid URL illegal constructor error

onsclom:
`nodeFileTrace` errors on `const URLParser = typeof window === 'undefined' ? URL : 'b'` with `TypeError: Class constructor URL cannot be invoked without 'new'`.

This is because this code:

```ts
async function ConditionalExpression() {
  return {
    test: 'foo',
    then: URL,
    else: 'bar',
  };
}
await ConditionalExpression();
```

calls `URL()` since `ConditionalExpression` is an async function returning a [thenable](https://masteringjs.io/tutorials/fundamentals/thenable).

Solution: rename all `.then` fields to `.ifTrue` since `then` is reserved for thenables.

Closes https://github.com/vercel/nft/pull/447

### DIFF
--- a/src/test.ts
+++ b/src/test.ts
@@ -1,3 +1,0 @@
-import { nodeFileTrace } from './node-file-trace';
-
-nodeFileTrace(['./test/unit/a-url-error/input.js']);
--- a/src/test.ts
+++ b/src/test.ts
@@ -1,0 +1,3 @@
+import { nodeFileTrace } from './node-file-trace';
+
+nodeFileTrace(['./test/unit/a-url-error/input.js']);
--- a/src/utils/static-eval.ts
+++ b/src/utils/static-eval.ts
@@ -16,7 +16,7 @@ export async function evaluate(
 
   // walk returns:
   // 1. Single known value: { value: value }
-  // 2. Conditional value: { test, then, else }
+  // 2. Conditional value: { test, ifTrue, else }
   // 3. Unknown value: undefined
   function walk(node: Node) {
     const visitor = visitors[node.type];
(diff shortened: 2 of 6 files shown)


src/test.ts
styfle, line 1: I think this file should be removed, right?

src/utils/static-eval.ts
```diff
@@ -327,7 +327,7 @@ const visitors: Record<
 
     return {
       test: node.test,
-      then: thenValue.value,
+      then: thenValue.value === URL ? undefined : thenValue.value,
```
styfle, line 330: This looks wrong. Why would URL always be undefined? 

I suspect the root cause is somewhere else since this code represents the ternary from what I can tell.

TooTallNate, line 330: This will be confusing for future readers. Can we add a comment explaining why the special case here?

TooTallNate (changes requested):

src/utils/types.ts
```diff
@@ -8,7 +8,8 @@ export interface StaticValue {
 
 export interface ConditionalValue {
   test: string;
-  then: any;
+  // then is reserved for thenables
+  ifTrue: any;
   else: any;
 }
```
onsclom, line 14: Went with a new (and I think better) solution!

`then` is reservered for thenables, so changed everything in the code from `then` -> `ifTrue`.

styfle (approved):
Wow, nice catch!

src/utils/types.ts
```diff
@@ -8,7 +8,8 @@ export interface StaticValue {
 
 export interface ConditionalValue {
   test: string;
-  then: any;
+  // then is reserved for thenables
```
styfle, line 11: I think you can remove the comment. Putting this comment in the PR description should be sufficient.

TooTallNate (approved):

PR #459, opened 2024-12-25
feat: support `module.createRequire`

### LINKED ISSUE #430: Does not resolve dependencies loaded via `module.createRequire`
The current recommended way to load native node modules from ESM is via `module.createRequire` but `@vercel/nft` does not resolve and include paths that are loaded this way:
```ts
import * as module from 'node:module';

const req = module.createRequire(import.meta.url);
const lib = req('./lib.node');
```

maxam2017:
## Description
Added support for tracing dependencies loaded via `module.createRequire`.

Fixes #430

### DIFF
--- a/src/analyze.ts
+++ b/src/analyze.ts
@@ -1082,6 +1082,20 @@ export default async function analyze(
           if (returned) setKnownBinding(fnName.name, { value: BOUND_REQUIRE });
         }
       }
+      // Support module.createRequire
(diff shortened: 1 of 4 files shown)


ijjk (approved):
This makes sense to me to support, thanks for the PR!

styfle (approved):
Great, thanks! 🎉

Issue #457, opened 2024-12-13
0.27.8 requires node `^18.17.0 || >=20.5.0` but nft intends to support node>=16


serhalp:
👋🏼 Hi folks!

[This PR to upgrade node-pre-gyp](https://github.com/vercel/nft/pull/453) (released in 0.27.8) indirectly [bumped a transitive dependency `nopt`](https://github.com/vercel/nft/pull/453/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519R3605) from 5.0.0 to 8.0.0.

and [nopt@8.0.0](https://github.com/npm/nopt/releases/tag/v8.0.0) requires node `^18.17.0 || >=20.5.0`.

However, `@vercel/nft` has `package.json#engines#node` of `>=16`: https://github.com/vercel/nft/blob/7fbb55944f11a1b2d950de479992c2cf3c49a574/package.json#L145-L147

TL;DR: nft intends to support node 16+ but no longer supports 16 to 18.16.0

I realize since nft is still on 0.x it can introduce breaking changes at any time, but I'm guessing this wasn't intentional! If it is, could we update the `engines.node` and note this in the release notes?

Thanks! Happy to open a PR if you all know what direction you want to take.

styfle:
engines.node should be updated since node@16 has reached EOL a year ago

https://nodejs.org/en/blog/announcements/nodejs16-eol

serhalp:
@styfle would you accept a PR to revert this change in another patch (0.27.9) and a PR to update `engines.node` to `>=18` (new minor? 0.x semantics are weird...)

styfle:
Sorry about the delay.

I checked when we dropped support for node@16 and it was a year ago in PR https://github.com/vercel/nft/pull/374 

Since that bumped the semver version correctly, I think we can make a new PR with a patch to bump `engines` to match that.

serhalp:
That makes sense, thanks! Here's a PR: https://github.com/vercel/nft/pull/461.

PR #461, opened 2025-01-02
fix: require node 18 in package.json

serhalp:
As discussed [here](https://github.com/vercel/nft/issues/457#issuecomment-2568065889), node 16 support was already dropped in 0.26.0, but `package.json#engines#node` was not updated.

Closes https://github.com/vercel/nft/issues/457.

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,7 +143,7 @@
   },
   "packageManager": "npm@10.2.5",
   "engines": {
-    "node": ">=16"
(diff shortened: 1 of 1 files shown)


styfle:
Thanks! I made a PR at the same time and that also bumps tsconfig.json so I'll close this one

- https://github.com/vercel/nft/pull/460

PR #462, opened 2025-01-26
fix: bump @mapbox/node-pre-gyp

talentlessguy:
@mapbox/node-pre-gyp just released their version 2 a few days ago, this pull request simply bumps rc to stable v2

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0-development",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


EndangeredMassa (approved):
It looks like there's no real change between these versions.

`node-pre-gyp` does say it doesn't support Node 16, which we still need to do, but the RC we're currently using already said that. So, I think this is a fine change.

Thanks!

styfle (approved):

PR #491, opened 2025-02-18
chore: Remove deprecated library polyfill-library

### LINKED ISSUE #490: Remove deprecated package polyfill-library
Remove deprecated package polyfill-library

thomashohn:
Resolves #490


thomashohn:
The package was removed - so whats that correct thing to do then?

styfle:
`@vercel/nft` is used by Vercel deployments and therefore could be used to trace deprecated packages.

We can't introduce breaking behavior and dropping a test would risk that.

thomashohn:
@styfle Readded test - but uncertain if the test won't fail now.
Maybe we should skip this PR and merge some of the other PR's?

test/integration.test.js
```diff
@@ -41,6 +41,11 @@ for (const integrationTest of filteredTestsToRun) {
     const fails = integrationTest.endsWith('failure.js');
     let traceBase = path.resolve(__dirname, '..');
 
+    if (integrationTest === 'polyfill-library.js') {
+      console.log('Skipping polyfill-library.js');
+      return;
+    }
```
styfle, line 47: Not sure I understand this change. It has the same problem - you're not testing the polyfill-libarary.

Lets close this PR.

PR #521, opened 2025-03-24
chore: Bump dependencies

thomashohn:
Update deps


thomashohn:
@styfle Ready for review

package.json
```diff
@@ -31,7 +31,7 @@
     "estree-walker": "2.0.2",
     "glob": "^10.4.5",
     "graceful-fs": "^4.2.9",
-    "node-gyp-build": "^4.2.2",
+    "node-gyp-build": "~4.7.1",
```
styfle, line 34: `^4.2.2` already covers all of 4.x because of the upcaret

Why is this being changed to a tilde?

package.json
```diff
@@ -53,7 +53,7 @@
     "@types/picomatch": "^3.0.1",
     "@vercel/git-hooks": "^1.0.0",
     "@vercel/style-guide": "^5.2.0",
-    "analytics-node": "^3.4.0-beta.1",
+    "analytics-node": "^3.5",
```
styfle, line 56: Latest is 6.2.0, lets use that instead

package.json
thomashohn: Could be that the 4.8 did break stuff - will have a second look

package.json
thomashohn: Will have a look at it - but switch from 3 to 6 might have been breaking stuff

package.json
```diff
@@ -31,7 +31,7 @@
     "estree-walker": "2.0.2",
     "glob": "^10.4.5",
     "graceful-fs": "^4.2.9",
-    "node-gyp-build": "^4.2.2",
+    "node-gyp-build": "^4.8.4",
```
styfle, line 34: I think this should be a separate PR since its not a `chore:`, its a `fix:`

(changes to `dependencies` affect consumers)

package.json
thomashohn: @styfle - reverted

package.json
thomashohn: Done

package.json
thomashohn: Reverted

thomashohn:
@styfle - ready to rock

styfle (approved):

PR #534, opened 2025-07-22
feat: add support for module-sync export condition in Node.js 22+

### LINKED ISSUE #531: Support for "module-sync" exports condition
Hello, we are currently using `@vercel/nft` to collect dependencies for our application. Thanks the capabilities it provides, which have saved us a significant amount of server resources and deployment time."

But recently we encountered a problem. When a package uses `module-sync` exports, some unexpected behaviors occur.

When using higher versions of Node 22, the exports of `module-sync` are used at runtime. However, the collected dependencies are `require`, `node` or `default`. These exported files may not be consistent with `module-sync`, so running with the collected dependencies might lead to issues where files cannot be found.

---

I would like to know if `@vercel/nft` will support the `module-sync` exports condition. If there are plans, I can try to provide a PR. 

However, what I am currently unsure about is whether we should consider the situation where the build and deployment are on different Node versions. For example, the build uses Node 22 while the deployment uses Node 20. Although this is not a correct practice, it indeed exists widely. If we need to consider this difference, perhaps a configuration option should be provided, or there might already be an existing option that I am unaware of but can adapt to this scenario.

Please correct me if I have any misunderstandings.

hamidrezahanafi:
This PR implements support for the module-sync export condition in package.json exports, which is available in Node.js 22 and later versions. The feature enables packages to provide synchronous ES module loading behavior for enhanced performance in supported Node.js versions.
Unit tests added with `process.versions.node` mocked

- Fixes #531


test/unit.test.js
```diff
@@ -82,6 +83,24 @@ for (const { testName, isRoot } of unitTests) {
   const unitPath = join(__dirname, 'unit', testName);
 
   it(`should correctly trace ${testSuffix}`, async () => {
+    // Mock Node.js version for module-sync-condition tests
+    if (testName === 'module-sync-condition') {
```
styfle, line 87: Instead of mocking, lets just skip for older versions of node since we have a ci matrix already.

See the `continue` statements above.

src/resolve-dependency.ts
```diff
@@ -168,6 +169,9 @@ function getExportsTarget(
         condition === 'default' ||
         (condition === 'require' && cjsResolve) ||
         (condition === 'import' && !cjsResolve) ||
+        (condition === 'module-sync' &&
+          getNodeMajorVersion() >= 22 &&
+          !cjsResolve) ||
```
styfle, line 174: Are you sure about the `!cjsResolve` here?

The docs say

> "module-sync" - matches no matter the package is loaded via import, import() or require()

https://nodejs.org/api/packages.html#conditional-exports

test/unit.test.js
styfle: Now that I think about it, we need this first:

- https://github.com/vercel/nft/pull/535

src/resolve-dependency.ts
hamidrezahanafi: I added because of the what's next come in the doc
>The format is expected to be ES modules

I might be wrong though

test/unit/module-sync-condition-node20/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "test-pkg-sync-fallback",
+  "type": "module",
```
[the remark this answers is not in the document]
hamidrezahanafi: Okay tested again, tests are failing with `cjs` but it really doesn't need the check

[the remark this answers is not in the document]
styfle: I can't remember if `cjeResolve` is implying the current module was parsed as cjs or if its the module being imported. Thats why it might need to be unconditional.

Regardless, we should have a test for both types since that works with Node.js as far as I can tell.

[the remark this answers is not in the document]
styfle: Done

styfle (approved):
Great work, thanks!

PR #542, opened 2025-09-24
fix: fs.readFile emit relative assets using cwd

icyJoseph:
In this input file:

```js
const fs = require('fs');

fs.readFile('./asset1.txt')

fs.readFile(require.resolve('./asset2.txt'))

const _basePath = __dirname;
const asset3 = 'asset3.txt';
fs.readFileSync(_basePath + '/' + asset3, 'utf8');
```

We don't emit for `fs.readFile('./asset1.txt')`, but we should:

> Regardless of where the code is (i.e. in which chunk it ends up in after bundling), readFile is always relative to process.cwd() which is set by Next.js to always be the project root (otherwise process.cwd() + "./foo" wouldn't work either


src/analyze.ts
```diff
@@ -1196,6 +1196,14 @@ export default async function analyze(
       return;
     }
 
+    if (
+      'value' in staticChildValue &&
+      typeof staticChildValue.value === 'string' &&
+      staticChildValue.value.startsWith('.')
+    ) {
+      staticChildValue.value = path.resolve(dir, staticChildValue.value);
```
ijjk, line 1204: Are we matching behavior here? https://github.com/vercel/next.js/pull/84155 seems like this is diverging

icyJoseph: Mmm - yeah, I have it wrong here, I should use `cwd` - and update the test

icyJoseph: In this branch, running `npm run build && node out/cli.js print test/unit/fs-emission/input.js` won't output `asset1.txt` cuz it doesn't exist in the root. 

If we place a dummy `asset1.txt` at the root and re-run, then we trace the file.

src/analyze.ts
```diff
@@ -1196,6 +1196,14 @@ export default async function analyze(
       return;
     }
 
+    if (
+      'value' in staticChildValue &&
+      typeof staticChildValue.value === 'string' &&
+      staticChildValue.value.startsWith('.')
```
styfle, line 1202: Since this code is checking for `.`, I think we need a few more tests.

For example:

```js
readFile('../foo.txt')
readFile('../../bar.txt')
readFile('../bar/../baz.txt')
```

src/analyze.ts
ijjk: I think we have two variations one with `cwd` at the root and once in the fixture

src/analyze.ts
icyJoseph: Could we flip this though? The branch under checks `isAbsolutePathOrUrl`, if we fail that check, we could enter an `else if('value' in staticChildValue &&
      typeof staticChildValue.value === 'string)` branch where we try to resolve if `cwd` + the static value exists, and if so trace it.

test/unit/fs-relative-emission/main/input.js
```diff
@@ -0,0 +1,11 @@
...
+const fs = require('fs');
+
+fs.readFile('../asset1.txt')
+
+fs.readFile('../blog/asset2.txt')
+
+const asset3 = 'asset3.txt';
+fs.readFileSync('../blog/author' + '/' + asset3, 'utf8');
+
+fs.readFile('../blog/author/../asset4.txt')
```
styfle, line 11:
```suggestion
fs.readFileSync('../asset1.txt')

fs.readFileSync('../blog/asset2.txt')

const asset3 = 'asset3.txt';
fs.readFileSync('../blog/author' + '/' + asset3, 'utf8');

fs.readFileSync('../blog/author/../asset4.txt')
```

Small suggestion since this code will fail when running `node input.js` right now.

test/unit.test.js
```diff
@@ -90,7 +90,11 @@ for (const { testName, isRoot } of unitTests) {
     console.log(`Skipping unit test on Node.js 22 or above: ${testSuffix}`);
     continue;
   }
-  const unitPath = join(__dirname, 'unit', testName);
+  let unitPath = join(__dirname, 'unit', testName);
+
+  if (testName === 'fs-relative-emission') {
+    unitPath += '/main';
```
styfle, line 96:
```suggestion
    // this test has file reads that traverse to the parent dir
    // so we need to start from the child dir.
    unitPath += '/main';
```

Lets mention why this is here

styfle (approved):
Great work, thanks! 🥇

PR #550, opened 2025-11-18
feat: Ensure module-sync conditions also trace cjs fallback

jeffsee55:
fix: Include fallback condition when evaluating module-sync exports

Previously, when resolving the `module-sync` condition, we only traced the `module-sync` file itself, not the fallback (`require` or `default` condition). This caused runtimes with require(esm) disabled to be unable to find the module.

Also, fixed a bug in the `module-sync-condition-cjs` test where:
- The input was importing from `test-pkg-sync-es` instead of `test-pkg-sync-cjs` (wrong package name)
- The test was running on both Node 20 and Node 22, producing inconsistent results
- Changed input to use `require()` syntax (appropriate for CJS packages)
- Now only runs on Node 22+ where module-sync is supported
- Created separate `module-sync-condition-cjs-node20` test for Node 20 behavior

## Test Coverage

| Test | Node | Syntax | Files Traced | Reason |
|------|------|--------|--------------|--------|
| `module-sync-condition-cjs` | 22+ | `require()` | `module-sync.js`, `fallback.js` | Uses module-sync + fallback for require(esm) opt-out |
| `module-sync-condition-cjs-node20` | 20 | `require()` | `fallback.js` | Falls back to default (no module-sync support) |
| `module-sync-condition-es` | 22+ | `import` | `module-sync.js`, `fallback.js` | Uses module-sync + fallback for require(esm) opt-out |
| `module-sync-condition-es-node20` | 20 | `import` | `import.js` | Falls back to import condition (no module-sync support) |

---

Also adjusts jest thresholds, here are the main paths this PR added that it did not add test coverage for, mostly error handling

Screenshot 2025-11-20 at 11 57 37 AM

### DIFF
--- a/src/resolve-dependency.ts
+++ b/src/resolve-dependency.ts
@@ -185,14 +185,37 @@ function getExportsTarget(
   return undefined;
 }
 
-function resolveExportsImports(
+async function validateAndResolvePaths(
+  paths: string[],
+  parent: string,
+  job: Job,
+  cjsResolve: boolean,
+): Promise<string[]> {
+  const validatedPaths: string[] = [];
+  for (const path of paths) {
+    if (cjsResolve) {
+      const resolved =
+        (await resolveFile(path, parent, job)) ||
+        (await resolveDir(path, parent, job));
+      if (!resolved) throw new NotFoundError(path, parent);
+      validatedPaths.push(resolved);
+    } else {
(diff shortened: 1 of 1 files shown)


src/resolve-dependency.ts
styfle: I'm starting to think that maybe resolveExportsImports should just normalize to always return an array.

PR #558, opened 2025-12-23
feat: support `module.createRequire` when `mixedModules: false`

### LINKED ISSUE #543: Resolving `module.createRequire` throws error when `mixedModules: false`
Support for `createRequire` was added in:
- #459 

However, this only works with `mixedModules: true`. 

If this mode is not enabled, you get:

```
    TypeError: Cannot read properties of undefined (reading 'shadowDepth')

      623 |                                 node.arguments[0].type === 'Literal' &&
      624 |                                 node.callee.type === 'Identifier' &&
    > 625 |                                 knownBindings.require.shadowDepth === 0) {
          |                                                       ^
      626 |                                 await processRequireArg(node.arguments[0]);
      627 |                             }
      628 |                             break;

      at Object.shadowDepth (out/analyze.js:625:55)
      at AsyncWalker.visit (node_modules/estree-walker/dist/umd/estree-walker.js:234:6)
      at AsyncWalker.visit (node_modules/estree-walker/dist/umd/estree-walker.js:271:7)
      at AsyncWalker.visit (node_modules/estree-walker/dist/umd/estree-walker.js:264:15)
      at AsyncWalker.visit (node_modules/estree-walker/dist/umd/estree-walker.js:264:15)
      at asyncWalk (node_modules/estree-walker/dist/umd/estree-walker.js:336:10)
      at analyze (out/analyze.js:508:5)
      at Job.emitDependency (out/node-file-trace.js:340:29)
          at async Promise.all (index 0)
      at nodeFileTrace (out/node-file-trace.js:62:5)
      at doTrace (test/unit.test.js:153:37)
      at Object.<anonymous> (test/unit.test.js:342:5)
```

The error at this line is because `knownBindings.require` is undefined when the file is ESM and `mixedModules: false`:
https://github.com/vercel/nft/blob/b10961add98da050953cba70756ba16a6911c77e/src/analyze.ts#L734

This was not picked up buy the tests because it looks like they all have `mixedModules` enabled:
https://github.com/vercel/nft/blob/b10961add98da050953cba70756ba16a6911c77e/test/unit.test.js#L169

logaretm:
This PR adds several fixes and tests for various edge cases around `module.createRequire`.

### Null Checks

When `mixedModules: false` and the file is ESM, `knownBindings.require` is `undefined` because it's only initialized for CommonJS or when `mixedModules: true`. The code was accessing `knownBindings.require.shadowDepth` without checking if `require` exists first.

### Blocking variables named `require`

This came up by @timfish in [this comment](https://github.com/vercel/nft/issues/543#issuecomment-3333430729), we seem to actively block variables named `require`. 

This is marked as "known unknown" but I thought we should make an exception if the variable was created by `module.createRequire`.

### Named import tracking

Using:

```js
import { createRequire } from 'node:module'
```

wasn't being detected, so I added a detection block to support it.

closes #543

### DIFF
--- a/test/unit.test.js
+++ b/test/unit.test.js
@@ -145,7 +145,8 @@ for (const { testName, isRoot } of unitTests) {
         testName === 'module-create-require-no-mixed' ||
         testName === 'module-create-require-named-require' ||
         testName === 'module-create-require-named-import' ||
-        testName === 'module-create-require-ignore-other'
+        testName === 'module-create-require-ignore-other' ||
+        testName === 'module-create-require-destructure'
       ) {
         inputFileNames = ['input.mjs'];
       }
--- a/src/analyze.ts
+++ b/src/analyze.ts
@@ -1098,7 +1098,10 @@ export default async function analyze(
         node.callee.property.type === 'Identifier' &&
         node.callee.property.name === 'createRequire'
       ) {
-        if (parent.type === 'VariableDeclarator') {
+        if (
+          parent.type === 'VariableDeclarator' &&
+          parent.id.type === 'Identifier'
+        ) {
           const requireName = parent.id.name;
           setKnownBinding(requireName, { value: BOUND_REQUIRE });
         }
@@ -1115,7 +1118,10 @@ export default async function analyze(
           'value' in createRequireBinding &&
           createRequireBinding.value === CREATE_REQUIRE
         ) {
-          if (parent.type === 'VariableDeclarator') {
+          if (
+            parent.type === 'VariableDeclarator' &&
+            parent.id.type === 'Identifier'
+          ) {
             const requireName = parent.id.name;
             setKnownBinding(requireName, { value: BOUND_REQUIRE });
           }
--- a/test/unit/module-create-require-destructure-namespace/dep.js
+++ b/test/unit/module-create-require-destructure-namespace/dep.js
@@ -1,0 +1,1 @@
+module.exports = 'dependency';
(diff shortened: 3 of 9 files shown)


test/unit.test.js
```diff
@@ -176,7 +184,8 @@ for (const { testName, isRoot } of unitTests) {
           log: true,
           // disable analysis for basic-analysis unit tests
           analysis: !testName.startsWith('basic-analysis'),
-          mixedModules: true,
+          mixedModules:
+            testOpts.mixedModules !== undefined ? testOpts.mixedModules : true,
```
styfle:
```suggestion
            testOpts?.mixedModules ?? true,
```

test/unit/module-create-require-block-other/input.mjs
```diff
@@ -0,0 +1,7 @@
+import * as module from 'node:module';
+
+// This require variable is NOT from module.createRequire, so it should be blocked
```
styfle: What does "blocked" mean here?

logaretm: Maybe the wrong way to describe it, but I meant that it shouldn't register those calls which is the current behavior. So we analyze the `require` named variables through and only if they are created with `module.createRequire` but not the other random variables named `require`.

Does that make sense? should I re-word it differently?

logaretm: Yea, that makes more sense. I will update the test names and comments for clarity

PR #561, opened 2026-01-12
feat: add depth option

### LINKED ISSUE #559: Export `analyze` function and `Job` class
## Problem

Currently, only `nodeFileTrace` is exported, which always performs full recursive dependency tracing. There's no way to analyze a single file without tracing all its dependencies.

## Use Case

**Single file analysis in build tools**
```typescript
// Vite/Rollup plugin that needs per-file analysis
async transform(code, id) {
  const job = new Job(...);
  const result = await analyze(id, code, job);
  // ...
}
```

## Proposed Solution

Export `analyze` and `Job` from `src/index.ts`.

**Question:** Would this be sufficient for the use cases above, or are there limitations in the current implementation that would need to be addressed first?

magne4000:
- Closes #559

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -215,7 +215,7 @@ const { fileList } = await nodeFileTrace(files, {
 });
 ```
 
-With `depth: 0`, only the entry point files themselves are included in the trace, with no dependencies followed. With `depth: 1`, the entry points and their direct dependencies are included, and so on.
+With `depth: 0`, only the entry point files themselves are included in the trace, with no additional modules. With `depth: 1`, the entry points and their directly imported modules are traced, and so on.
 
 Note that the `depth` option cannot be less than 0. Setting a negative value will result in an error.
 


test/unit.test.js
```diff
@@ -193,6 +206,7 @@ for (const { testName, isRoot } of unitTests) {
           resolve: testName.startsWith('resolve-hook')
             ? (id, parent) => `custom-resolution-${id}`
             : undefined,
+          depth,
```
styfle: Lets revert `unit.test.js` changes and instead change this one line to use `depth: testOpts.depth` instead.

Then separate the 1 test fixture into 4 different test fixtures.

styfle (commented):
Readme needs to be updated too to mention the new option (and the caveat it can't be less than 0)

src/node-file-trace.ts
```diff
@@ -318,7 +322,12 @@ export class Job {
     return undefined;
   }
 
-  async emitDependency(path: string, parent?: string) {
+  async emitDependency(
+    path: string,
+    parent?: string,
+    depth: number = this.depth,
+  ) {
+    if (depth < 0) throw new Error('{ depth } must be positive.');
```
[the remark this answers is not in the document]
styfle: So this error needs to happen much early perhaps in `new Job()`, not in `emitDependency()`

src/node-file-trace.ts
```diff
@@ -331,7 +341,9 @@ export class Job {
     const additionalDeps = this.remappings.get(path);
     if (additionalDeps) {
       await Promise.all(
-        [...additionalDeps].map(async (dep) => this.emitDependency(dep, path)),
+        [...additionalDeps].map(async (dep) =>
+          this.emitDependency(dep, path, depth),
```
styfle:
```suggestion
          this.emitDependency(dep, path, depth - 1),
```

This recursive call should also decrement, right?

Or do we special case the remappings?

readme.md
```diff
@@ -205,6 +205,20 @@ const { fileList } = await nodeFileTrace(['index.ts'], { cache });
...
+
+The `depth` option controls how many levels of dependencies to trace from the entry point files. This can be useful for limiting the scope of the trace or for debugging purposes:
+
+```js
+const { fileList } = await nodeFileTrace(files, {
+  depth: 2,
+});
+```
+
+With `depth: 0`, only the entry point files themselves are included in the trace, with no dependencies followed. With `depth: 1`, the entry points and their direct dependencies are included, and so on.
```
styfle:
```suggestion
With `depth: 0`, only the entry point files themselves are included in the trace, with no additional modules. With `depth: 1`, the entry points and their directly imported modules are traced, and so on.
```

src/node-file-trace.ts
magne4000: I was thinking of handling remapping as same depth, as conceptually they do not add a new child/parent relationship, but they "replace" the current module.

PR #575, opened 2026-04-28
feat: handle external packages in imports field

mischnic:
https://nodejs.org/api/packages.html#subpath-exports:

> All target paths in the ["exports"](https://nodejs.org/api/packages.html#exports) map (the values associated with export keys) must be relative URL strings starting with `./`

but

>Unlike the "exports" field, the "imports" field permits mapping to external packages.
>The resolution rules for the imports field are otherwise analogous to the exports field.

### DIFF
--- a/src/resolve-dependency.ts
+++ b/src/resolve-dependency.ts
@@ -271,6 +271,10 @@ async function resolveExportsImports(
       }
 
       return await validateAndResolvePaths(paths, parent, job, cjsResolve);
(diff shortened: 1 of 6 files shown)


styfle (approved):
Interesting, I would expect that `test/unit/imports/node_modules/pkg/package.json` would be necessary to make this work but it seems its not required when running `node test/unit/imports/input.js`, so looks good to me, thanks!

PR #581, opened 2026-05-20
feat: module-sync fallback should use import for ESM

mischnic:
I found this while implementing the Turbopack version: https://github.com/vercel/next.js/pull/93970

When doing an ESM `import`, it used the `module-sync` condition (correct) and also included the `default` condition, even if there was a `import` condition (which is what older Node.js versions would have used)
```
  "exports": {
    ".": {
      "module-sync": "./module-sync.js", // included (correct)
      "import": "./import.js", // skipped (incorrect)
      "default": "./fallback.js" // included (incorrect)
    }
  }
```


jeffsee55 (approved):

src/resolve-dependency.ts
```diff
@@ -253,7 +253,11 @@ async function resolveExportsImports(
         getNodeMajorVersion() >= 22
       ) {
         const fallbackCondition =
-          'require' in exportsForSubpath ? 'require' : 'default';
+          cjsResolve && 'require' in exportsForSubpath
+            ? 'require'
+            : !cjsResolve && 'import' in exportsForSubpath
+              ? 'import'
+              : 'default';
```
styfle, line 260: Do we have tests covering all these cases?

mischnic: I added one more test, now we do.

styfle (approved):
Thanks! 🎉

PR #582, opened 2026-05-20
feat: module-sync fallback should use import for ESM

jeffsee55:
Updates module-sync fallback tracing so ESM imports include the import fallback rather than default when module-sync is selected.

Tested with the existing module-sync unit coverage in this branch.

Made with [Cursor](https://cursor.com)


jeffsee55:
Closing because this PR was opened from the upstream module-sync fallback branch commits rather than the local catchall changes.

PR #584, opened 2026-05-20
feat: add module-sync catchall tracing

jeffsee55:
## Summary
- Add a `moduleSyncCatchall` trace option to include both `module-sync` and normal fallback export targets.
- Preserve existing module-sync fallback behavior while extending catchall coverage to wildcard export patterns.
- Document the new option and add focused unit coverage.

## Test plan
- `npm run build`
- `npx prettier --check readme.md src/node-file-trace.ts src/resolve-dependency.ts src/types.ts test/unit.test.js test/unit/imports-module-sync/output.js test/unit/imports-module-sync-cjs/output.js test/unit/module-sync-condition-cjs/output.js test/unit/module-sync-condition-es/output.js test/unit/module-sync-condition-es-default/output.js test/unit/module-sync-condition-es-nested/output.js test/unit/self-reference-module-sync/output.js test/unit/module-sync-catchall/input.js test/unit/module-sync-catchall/package.json test/unit/module-sync-catchall/sync/feature.js test/unit/module-sync-catchall/fallback/feature.js test/unit/module-sync-catchall/output.js test/unit/module-sync-catchall/test-opts.json`
- `npx jest test/unit.test.js --runInBand --testNamePattern="module-sync"`

Made with [Cursor](https://cursor.com)


src/resolve-dependency.ts
```diff
@@ -239,6 +246,7 @@ async function resolveExportsImports(
       matchObj[subpath],
       job.conditions,
       cjsResolve,
+      getNodeMajorVersion() >= 22,
```
styfle: I'm thinking we should remove the node detection in nft and instead rely on the option being passed in.

We can have it false by default for now and then swap to true in the future.

src/resolve-dependency.ts
```diff
@@ -169,13 +175,14 @@ function getExportsTarget(
         condition === 'default' ||
         (condition === 'require' && cjsResolve) ||
         (condition === 'import' && !cjsResolve) ||
-        (condition === 'module-sync' && getNodeMajorVersion() >= 22) ||
-        conditions.includes(condition)
+        (condition === 'module-sync' && moduleSyncSelectable) ||
+        (condition !== 'module-sync' && conditions.includes(condition))
```
styfle: Is this `condition !== 'module-sync'` needed?

We didn't have it before 🤔

src/resolve-dependency.ts
styfle: Actually I think your solution is fine. Lets just hoist `getNodeMajorVersion() >= 22` up to the top so we can reuse the variable everywhere in this file instead of passing it around

styfle (approved):

PR #586, opened 2026-05-22
feat: refactor ffmpeg-static special case

styfle:
Previously, this code would do a dynamic require but now it does static analysis on the possible output files.

I had a clanker go through all possible outputs from all versions to come up with this list.

Then I simplified it into the two fs calls here.

image


styfle:
Note - I marked this as `feat:` instead of `fix:` since there might be some risk in this change - the code is quite dynamic and its possible something was missed.

ijjk (approved):

PR #590, opened 2026-05-27
fix: lock semantic-release publish dependencies

styfle:
We have a problem where semantic-release is only called in ci.yml but it doesnt have a lockfile which means transitive dependencies could change.

I think the reason it was set up this way is to avoid slowing down CI for PRs that don't publish since they wont need to install semantic-release (40MB). Only main needs to publish.

The fix is to add a new publish directory with its own lock file, then we can install and invoke the bin from the publish directory to ensure dependencies are locked.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,11 +75,8 @@ jobs:
         env:
           BULL_REDIS_CONNECTION: ${{ (matrix.os == 'ubuntu-latest') && 'redis://127.0.0.1:6379/0' || '' }}
         run: pnpm ${{ (matrix.os == 'ubuntu-latest') && 'test-verbose' || 'test' }}
-      - name: Install Publish Dependencies
-        if: matrix.os == 'ubuntu-latest' && matrix.node == 22 && github.event_name == 'push' && github.ref == 'refs/heads/main'
-        run: pnpm --dir publish install --ignore-workspace --frozen-lockfile
       - name: Maybe Release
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -75,8 +75,11 @@ jobs:
         env:
           BULL_REDIS_CONNECTION: ${{ (matrix.os == 'ubuntu-latest') && 'redis://127.0.0.1:6379/0' || '' }}
         run: pnpm ${{ (matrix.os == 'ubuntu-latest') && 'test-verbose' || 'test' }}
+      - name: Install Publish Dependencies
+        if: matrix.os == 'ubuntu-latest' && matrix.node == 22 && github.event_name == 'push' && github.ref == 'refs/heads/main'
+        run: pnpm --dir publish install --ignore-workspace --frozen-lockfile
```
styfle:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -75,8 +75,11 @@ jobs:
...
       - name: Maybe Release
         if: matrix.os == 'ubuntu-latest' && matrix.node == 22 && github.event_name == 'push' && github.ref == 'refs/heads/main'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: pnpm dlx semantic-release@25.0.3
+        run: ./publish/node_modules/.bin/semantic-release
```
styfle:
```suggestion
        run: pnpm --dir publish install --ignore-workspace --frozen-lockfile && ./publish/node_modules/.bin/semantic-release
```

samcx (approved):

ijjk (approved):

PR #594, opened 2026-06-12
chore: upgrade sharp to latest v0.35.3

lovell:
~~Extends the sharp "special case" with additional ESM/CJS-specific paths added in v0.35.0.~~

No longer requires the "special case" from 0.35.2 onwards.

- https://github.com/lovell/sharp/pull/4509
- https://github.com/lovell/sharp/issues/4543

The existing path check remains to support older versions ~~and future-proof for when we switch to ESM-only~~.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     "semver": "^7.5.2",
     "sequelize": "^6.29.0",
     "serialport": "^13.0.0",
-    "sharp": "^0.34.3",
+    "sharp": "^0.35.3",
     "shiki": "^0.14.5",
     "socket.io": "^2.4.0",
     "socket.io-client": "^2.2.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -307,8 +307,8 @@ importers:
         specifier: ^13.0.0
         version: 13.0.0
       sharp:
-        specifier: ^0.34.3
-        version: 0.34.3
+        specifier: ^0.35.3
+        version: 0.35.3(@types/node@25.7.0)
       shiki:
         specifier: ^0.14.5
         version: 0.14.7
@@ -832,8 +832,8 @@ packages:
     resolution: {integrity: sha512-F9rL9k9Xjf5blCz8HsJRO4diy111cayL2vkY2XE4r4t3n0yPXVYy3KD3nJ1qbrSn9743UWSXH4IwuCa/HWlGFw==}
     engines: {node: '>=6.0.0'}
 
-  '@emnapi/runtime@1.4.5':
-    resolution: {integrity: sha512-++LApOtY0pEEz1zrd9vy1/zXVaVJJ/EbAF3u0fXIzPJEDtnITsBGbbK0EkM72amhl/R5b+5xx0Y/QhcVOpuulg==}
+  '@emnapi/runtime@1.11.1':
+    resolution: {integrity: sha512-vgj7R3y3Wgx24IQaGPA/R6YFXLHVMOZ0uVEyIQPaWs+rd1AzfEMXlAC22FYwO1XkKR6NPsq7mUandH8oIRdZFw==}
 
   '@esbuild/aix-ppc64@0.25.0':
     resolution: {integrity: sha512-O7vun9Sf8DFjH2UtqK8Ku3LkquL9SZL8OLY1T5NZkA34+wG3OQF7cl4Ql8vdNzM6fzBbYfLaiRLIOZ+2FOCgBQ==}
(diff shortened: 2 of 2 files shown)


styfle:
@lovell I wouldn't expect the special case to be required now that sharp fixed the logic to be statically analyzable here:

- https://github.com/lovell/sharp/issues/4380#issuecomment-4017526484

I think we should figure out the failure mode instead of changing the special case (the special case should just be for old versions of sharp)

lovell:
The prebuilt sharp packages e.g. `@img/sharp-linux-x64` are now correctly discovered, however the prebuilt libvips packages e.g. `@img/sharp-libvips-linux-x64` are not.

For a file to be considered necessary, is it enough to reference that file in the `exports` of a `package.json` file or does it have to be part of a `require` or `import` statement in code?

styfle:
Typically it needs to be part of a require() or import or readFile() to be statically analyzed and emitted, but there is a bit more nuance.

For example this sharedLibEmit() does some magic that might be relevant here: https://github.com/vercel/nft/blob/7e43ba44b3c08729882ac81748c9a28fadd8ab9d/src/utils/sharedlib-emit.ts#L20

styfle:
I had claude try a few different solutions and they were pretty verbose since it tried to parse the `.node` binary to find the dependencies.

One of the solutions was much more simple seen here:

- https://github.com/vercel/nft/pull/595

However, it feels relatively similar to the special case we already have for sharp here (checking package.json for optionalDependencies), so I'm thinking we might as well merge your PR since its less risk.

test/integration.test.js
```diff
@@ -64,7 +64,7 @@ for (const integrationTest of filteredTestsToRun) {
         path.join(tmpdir, 'package.json'),
         JSON.stringify({
           packageManager: 'pnpm@8.14.3',
-          dependencies: { sharp: '0.33.2' },
+          dependencies: { sharp: '0.35.1' },
```
styfle: Lets rever this to the old version so we still have coverage for the old version

lovell:
How about a possible alternative where the `@img/sharp-libvips-*` packages (for macOS and Linux) export an empty/stub `.node` file to make the dependency a bit more concrete?

For example, the logic in `@img/sharp-linux-x64` could be modified to include something like:
```js
try { require('@img/sharp-libvips-linux-x64/empty.node') } catch {}
```

This might be enough for the existing `sharedLibEmit` logic to then discover the other shared libraries, assuming `require` errors are ignored.

styfle:
@lovell Yes that would be great!

lovell:
I've published a release candidate v0.35.2-rc.2 of sharp with the `stub.node` approach and it appears to work with the existing `sharedLibEmit` logic and doesn't require the special case handling.

Marking this PR as draft for now until after v0.35.2 is released and the 2-day new package cool-down has passed.

I note the logic in TurboPack is slightly different from the logic in this package as it only has an `ends_with` check rather than a more comprehensive regex. I don't think this will work with common filenames like `shared.so.1.2.3`.

https://github.com/vercel/nft/blob/8673d639da0064ae02fa05d67b981100f6e0fbe1/src/utils/sharedlib-emit.ts#L16

https://github.com/vercel/next.js/blob/c5a826f59d9385363774c292d420bdfb54f02ec2/turbopack/crates/turbopack-resolve/src/node_native_binding.rs#L147

styfle:
I spoke to the @mischnic on the turbopack team we think require('pkg/stub.node') is probably going to be prone to error.

If you know the exact files that need to be included, you could instead do  `path.join(__dirname, 'file.so'), `path.join(__dirname, 'header.h')`, etc for the files you expect to be included.

You can test this locally without needing to publish a new version by using

```
npm i -g @vercel/nft
echo "require('path/to/sharp')" => example.cjs
nft print example.cjs
```

lovell:
I've now updated this PR to the latest sharp v0.35.3, which removes the stub and instead uses `require.resolve` to ensure the shared library binary is included in the list of files to include.

For example, the `index.cjs` file in the `@img/sharp-linux-x64` package contains:
```js
try { require.resolve('@img/sharp-libvips-linux-x64/binary'); } catch {}
```
...and the `package.json` file in the `@img/sharp-libvips-linux-x64` package contains:
```json
  "exports": {
    "./binary": "./lib/libvips-cpp.so.8.18.3"
  },
```
This approach works with the built-in (i.e. no special case) logic of the `@vercel/nft` package. Assuming `require.resolve` paths are traced it should also work with TurboPack and likely other bundlers.

styfle (approved):
Thanks for all your hard work on sharp!

styfle:
Upgrading in Next.js here: 

- https://github.com/vercel/next.js/pull/95507

PR #609, opened 2026-08-27
fix: do not glob from the filesystem root

Nic-Polumeyv:
A string like `` `/${dir}/routes` `` reaches `emitAssetPath` as `/<*>/routes`. On POSIX the parent is `''` and the stat fails. On Windows it resolves to `D:\<*>\routes`, the parent `D:` exists, and nft globs `D:/**/*/routes`, the entire drive.

The "do not emit directories above `__dirname`" rule in `validWildcard` only handled a trailing wildcard; it now checks the directory the wildcard sits in, which covers this and any other glob rooted at an ancestor of the importing file.

`test/wildcard-root.test.js` traces the fixture with `base` set to the filesystem root; the unit harness skips its `base: '/'` variant on Windows. Seen in the wild via sveltejs/kit#16963.

### DIFF
--- a/src/analyze.ts
+++ b/src/analyze.ts
@@ -1336,12 +1336,12 @@ export default async function analyze(
     if (assetPath.endsWith(path.sep + 'node_modules' + wildcardSuffix))
       return false;
     // do not emit directories above __dirname
-    if (
-      dir.startsWith(
-        assetPath.slice(0, assetPath.length - wildcardSuffix.length) + path.sep,
-      )
-    )
-      return false;
+    const wildcardIndex = assetPath.indexOf(WILDCARD);
+    const wildcardDir =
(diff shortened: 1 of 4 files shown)


ijjk (commented):
This is a breaking change. `base` and `ignore` already let callers define the permitted tracing boundary. A caller that sets `base` to the filesystem root is explicitly allowing files beneath that root, and may rely on inferred wildcards being traced there.

This patch adds a new global restriction based on the importing file’s directory, causing previously included files to be silently excluded despite being within `base`. The reported case should be handled by giving nft the intended project/common-ancestor `base`, or by excluding root-wide globs through `ignore`, as SvelteKit has done. I don’t think `validWildcard` should override existing caller configuration; otherwise this needs to be treated and documented as a breaking change.

